PR #5, opened 2025-11-05
Make mod show up in-game

MichaHelder:
- Makes the mod show up in the launcher
- Use old thumbnail (we should really make a new one)
- Gets rid of the intermediary MEIOUandTaxes folder as the game can't read from an extra folder deep like in EU4 it could


Dev2150 (approved):
Useful changes, thank you

PR #6, opened 2025-11-06
feat(societal_values): Make societal values behave like estate satisfaction equilibrium

Dev2150:
e.g. there is 75 Aristocracy & a trend towards Aristocracy of 0.5.

In vanilla, it stays at 75, until trend has different sign. 
If it's less than that, it will increase to 50. 
If it's more than 50, it maintains.

With this change, it will gravitate towards 50. 
If a negative event brings it to 40, it will slowly go back to 50.


in_game/common/scripted_effects/scripted_effects_societal_values_feedback_loop.txt
```diff
@@ -0,0 +1,39 @@
...
+    societal_value_balance = { SV_LEFT = capital_economy SV_RIGHT = traditional_economy }
+    societal_value_balance = { SV_LEFT = individualism SV_RIGHT = communalism }
+    societal_value_balance = { SV_LEFT = outward SV_RIGHT = inward }
+    societal_value_balance = { SV_LEFT = sinicized SV_RIGHT = unsinicized }
+    societal_value_balance = { SV_LEFT = absolutism SV_RIGHT = liberalism }
+    societal_value_balance = { SV_LEFT = mysticism SV_RIGHT = jurisprudence }
+}
+
+societal_value_balance = {
+    every_country = {
```
LuuBluum, line 21: Does this need `every_country` given that this event is a monthly country pulse? Doesn't this do `every_country` _for_ every country?

cheetawilliam (changes requested):
I agree with the idea, but for content to be accesible, such as turning into a republic, some goverment reforms there are not enough things that push you one way or the other. Those need to be looked at, requirements lowered or push factors added unless a lot of content becomes unaccasible.

Malorn-Deslor:
The thing is, it seems like society values are not really pulled to the center realistically. It's entirely possible for a real-world nation to sit at one extreme or the other.

PR #7, opened 2025-11-06
feat(societal_values): Make societal value behave like estate satisfaction equilibrium

Dev2150:
e.g. there is 75 Aristocracy & a trend towards Aristocracy of 0.5.

In vanilla, it stays at 75, until trend has different sign. 
If it's less than that, it will increase to 50.
If it's more than 50, it maintains.

With this change, it will gravitate towards 50.
If a negative event brings it to 40, it will slowly go back to 50.

### DIFF
--- a/in_game/common/on_action/on_actions_CENSUS.txt
+++ b/in_game/common/on_action/on_actions_CENSUS.txt
@@ -1,0 +1,5 @@
+﻿monthly_country_pulse = {
+    events = {
+        societal_values_balancing_feedback_loop.01
+    }
+}
--- a/in_game/common/scripted_effects/societal_values_feedback_loop_scripted_effects.txt
+++ b/in_game/common/scripted_effects/societal_values_feedback_loop_scripted_effects.txt
@@ -1,0 +1,39 @@
+﻿iterate_societal_values = {
+    societal_value_balance = { SV_LEFT = centralization SV_RIGHT = decentralization }
+    societal_value_balance = { SV_LEFT = traditionalist SV_RIGHT = innovative }
+    societal_value_balance = { SV_LEFT = spiritualist SV_RIGHT = humanist }
+    societal_value_balance = { SV_LEFT = aristocracy SV_RIGHT = plutocracy }
+    societal_value_balance = { SV_LEFT = serfdom SV_RIGHT = free_subjects }
+    societal_value_balance = { SV_LEFT = mercantilism SV_RIGHT = free_trade }
+    societal_value_balance = { SV_LEFT = belligerent SV_RIGHT = conciliatory }
+    societal_value_balance = { SV_LEFT = quality SV_RIGHT = quantity }
+    societal_value_balance = { SV_LEFT = offensive SV_RIGHT = defensive }
+    societal_value_balance = { SV_LEFT = land SV_RIGHT = naval }
+    societal_value_balance = { SV_LEFT = capital_economy SV_RIGHT = traditional_economy }
+    societal_value_balance = { SV_LEFT = individualism SV_RIGHT = communalism }
+    societal_value_balance = { SV_LEFT = outward SV_RIGHT = inward }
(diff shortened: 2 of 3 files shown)


cheetawilliam (changes requested):
I agree with the idea, but for content to be accesible, such as turning into a republic, ending disasters, some goverment reforms there are not enough things that push you one way or the other. Those need to be looked at, requirements lowered or push factors added unless a lot of content becomes unaccasible.

Malorn-Deslor:
So the thing is, I'm not sure they SHOULD be pushed to the middle. Realistically nations seem quite comfortable staying at extremes. A decentralized country doesn't really drift towards centralization. One doesn't have to put in effort to remain decentralized.

There are maybe three values which would logically 'drift' but they are one-sided. It does take effort to remain innovative, for example. Or to remain centralized, potentially. Or MAAAYBE to remain with free subjects, but that's a stretch.

But other things are historically traps, which would not drift towards the center. Serfdom, economic values, etc. Also there appears to already be some caps to how far you can go given a certain amount of drift, so we don't have the problem of all values maxxing out in every game.

Dev2150:
> I agree with the idea, but for content to be accesible, such as turning into a republic, ending disasters, some goverment reforms there are not enough things that push you one way or the other. Those need to be looked at, requirements lowered or push factors added unless a lot of content becomes unaccasible.

You mentioned disasters requiring 90 societal value (e.g. humanist). 
There is also a big requirement to change the government type.
In both cases, it requires a serious investment in laws & estate privileges and perhaps some luck from events

Dev2150:
> So the thing is, I'm not sure they SHOULD be pushed to the middle. Realistically nations seem quite comfortable staying at extremes. A decentralized country doesn't really drift towards centralization. One doesn't have to put in effort to remain decentralized.
> 
> There are maybe three values which would logically 'drift' but they are one-sided. It does take effort to remain innovative, for example. Or to remain centralized, potentially. Or MAAAYBE to remain with free subjects, but that's a stretch.
> 
> But other things are historically traps, which would not drift towards the center. Serfdom, economic values, etc. Also there appears to already be some caps to how far you can go given a certain amount of drift, so we don't have the problem of all values maxxing out in every game.

I see.
My motivation is that I do not like the arcadey strategy of treating trends like a resource: "Innovativess is 70 and the trend is 0.4, I don't see how I can improve it through the trend, so we can spend it on a favourable law giving aristocracy". 
Why would a country care as much about a value if you add a source in the opposite direction?
If you as a country care so much about quality, make the trend be 1 towards it and protect it. 
If it is 0, the country will become indifferent after hundreds of years.

Second of all, it's harder to have a strong value, so this change will strip countries of the society value bonuses, if they don't start with a decent trend. 
Thus, in this version, you can appreciate their benefits more.

Lastly, the trend speed can be tweaked.
The current speed factor is 0.01.

Assuming a maximum value & the trend = 0, this is how the value will change over time, assuming there are no events:

| Year | Speed factor = 0.005 | Speed factor = 0.010 |
| :------: | :----------: | :----: |
| 0   | 100 | 100 |
| 10 | 95 | 90 |
| 25 | 88 | 77 |
| 50 | 77 | 60 |
| 100 | 61 | 37 |
| 200   | 37 | 13 |
| 300 | 22 | 5 |

Malorn-Deslor:
> I see. My motivation is that I do not like the arcadey strategy of treating trends like a resource: "Innovativess is 70 and the trend is 0.4, I don't see how I can improve it through the trend, so we can spend it on a favourable law giving aristocracy". Why would a country care as much about a value if you add a source in the opposite direction? If you as a country care so much about quality, make the trend be 1 towards it and protect it. If it is 0, the country will become indifferent after hundreds of years.

But that's the thing, people won't become indifferent, that makes no sense. People tend to resist change of this sort. I agree that the caps and max should be made more organic, but I don't think a decay is the way to do that.

CorlisFataneya:
The discussion should be in an Issue NOT in a PR.

Issue #13, opened 2025-11-10
Impressing merchant vessels


VineFynn:
Idea is that early on your navy would have cogs/early transports as a decent part of your fleet, rather than just fishing boats with no offensive capability. For balance, this would require that burghers be crewing the ships, but at time of writing there's a bug preventing naval levies from using burghers as crew. So, putting a pin in this one

VineFynn:
Okay, the bug is still there but I've worked around it. Impressing merchant vessels is a go!

VineFynn:
Presumably we will want to be impressing later merchant vessels as well, but this is just for early-game balance

PR #14, opened 2025-11-10
Changes to boat levies

VineFynn:
Changes to boat levies:
Adds cogs as early-game naval levies from burghers living in a location with a wharf
Moves fishing boat levies to fishing rgos (and levies them from burghers)
closes https://github.com/MEIOU-and-Taxes/MnT-EU5/pull/14


Malorn-Deslor:
So, do we actually get to commit things now, is the question.

VineFynn:
> So, do we actually get to commit things now, is the question.

Yeah, though I don't mind if this sits here for a while, just wanted to submit it

Dev2150:
I see you only changed **ship_building_advance** in advances/0_age_of_traditions
It is sufficient to include that
Same logic for the other files

VineFynn:
> I see you only changed **ship_building_advance** in advances/0_age_of_traditions It is sufficient to include that Same logic for the other files

Hrm, I think this needs to be discussed in the discord

Dev2150 (approved):
OK 
image
image
image

VineFynn:
Close https://github.com/MEIOU-and-Taxes/MnT-EU5/issues/13

PR #15, opened 2025-11-10
Malorn balance changes to fix economy.

Malorn-Deslor:
Added lumber mill scrapwood production method that allow some wood production without tools.

Removed control penalty to pop promotion.


CorlisFataneya (approved):
Code is fine. As far as I can judge as of now. 
Maybe you could also add a solution to the meso america tool problems? :)

PR #19, opened 2025-11-10
Mapmodes for potential taxbase and enabled river mapmode

MateuszGie:
Two changes - one makes River mapmode in its current implementation available from the mapmode selection, second adds two mapmodes requested in suggestions thread, showing potential taxbase per location.

Clean error log


Dev2150:
I have a strong impression the entirety of map_modes.txt is not necessary. 
Have you checked if it works with a part of it & with a renamed file?

MateuszGie:
> I have a strong impression the entirety of map_modes.txt is not necessary. Have you checked if it works with a part of it & with a renamed file?

definitely what is needed is the first section with "@" values (constants?), without them we have those errors:

[pdx_persistent_reader.cpp:289]: Error: "Malformed token: @zoom_step_near, near line: 49" in file: "gfx/map/map_modes/mnt_map_modes.txt" near line: 49

if I add this section and keep two new and one modified mapmode, I get this: [gamedatabase.h:391]: Duplicated key rivers will not be created from file: gfx/map/map_modes/mnt_map_modes.txt:130 , which was sorted by using REPLACE. New commit pushed.

Dev2150 (approved):
I am happy hotkey assigment potentials are created
image
I have tested it and it works!
I believe it would be more fitting if it shares this kind of tooltip on hover, perhaps in a future iteration
image
It would be better if localizations are created for the new map modes

Dev2150 (changes requested):
I have tested it and it works!
I believe it would be more fitting if it shares this kind of tooltip on hover, perhaps in a future iteration
image

I am happy hotkey assigment potentials are created
image
It would be better if localizations work

image

Error logs say the localization does not have BOM

MateuszGie:
> I have tested it and it works! I believe it would be more fitting if it shares this kind of tooltip on hover, perhaps in a future iteration 

This nice panel may not be possible, it is scripted gui element that can be found at shared\location_tooltips.gui, line 2509. I will look into something simpler though.

> Error logs say the localization does not have BOM

This I don't understand, [loc strings are same as for vanilla map modes](https://github.com/MEIOU-and-Taxes/MnT-EU5/pull/19/files#diff-afbe9d2869cf9ded918a82645090177ccd20bd40f8f1e5f9ac2035a6410a1a61) and displayed properly:
image
and there is nothing in the logs.

Dev2150 (approved):
OK

PR #21, opened 2025-11-11
Add logs for total population, building type count & prices, for every region + Add PyQt6 plotter with filters & hotkeys

Dev2150:
To enable plotting, in April 1337, run in console: 
`event LOGGER_CHARTS.04`

When satisfied, run MT_grapher.py & follow the short instructions (assigning the path to the log folder)

There is always a month filter & a region filter. Sometimes there is a 3rd one.

There are 2 types of charts, based on the selected filters:
- If month filter is chosen, then there will be plotted a bar chart.
- Otherwise, it will be a line chart.

Tooltip is displayed, showing the 20 closest lines to the cursor (it wouldn't fit in the tooltip otherwise)
Hovered line is thickened for emphasis

Hotkeys:
- Press ` (backquote) to open table of the 3 charts with their corresponding hotkey
- When chart table is closed (e.g. when the app is started), the hotkeys (e.g. 1, 2, 3) can be pressed any time to switch between charts.
- Can press tab to quickly jump between UI elements
- When combobox is selected, can press up/down to quickly navigate between filters (e.g. months, regions, goods, building types)
- Can press right-click on combobox to reset filter
- Press F11 for full-screen

### DIFF
--- a/in_game/common/on_action/on_actions_CENSUS.txt
+++ b/in_game/common/on_action/on_actions_CENSUS.txt
@@ -1,0 +1,11 @@
+﻿on_monthly_pulse_country = {
+    events = {
+        LOGGER_CHARTS.02
+    }
+}
+
+on_game_started = {
+    events = {
+        LOGGER_CHARTS.03
+    }
+}
--- a/in_game/common/on_action/on_actions_CENSUS.txt
+++ b/in_game/common/on_action/on_actions_CENSUS.txt
@@ -11,8 +11,7 @@ on_game_start_MnT = {
 	}
 }
 
-
-monthly_country_pulse = {
+yearly_country_pulse = {
 	trigger = { c:FRA ?= this }
 	on_actions = {
 		once_per_month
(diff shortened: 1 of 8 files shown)


Dev2150:
image
image
image
image
image
image
image

Dev2150:
It now logs yearly, as discussed

Veheled (approved):
Very clean code and nice work! Nothing to complain from my side. 
Only problem with this is it would stop working when France stops existing i think. But to solve that we would need a hidden tag like in EU4 M&T, i think.

Dev2150:
> Very clean code and nice work! Nothing to complain from my side. Only problem with this is it would stop working when France stops existing i think. But to solve that we would need a hidden tag like in EU4 M&T, i think.

thank you! Yes, it would not work on 1st of Januaries where France does not exist, however I like to think this is very unlikely.
I don't know how that can be done, however I can improve it in a further iteration if needed

PR #22, opened 2025-11-11
Mepper branch merge; see desc for changes

MichaHelder:
- Don't let new born pops all be primary culture in cores, that just makes no sense. In vanilla it would automatically let _all_ babies in cores be of primary culture. Historically and realistically this is ridiculous, that is not how this works. 
- Add a readme.md in the main folder, extremely barebones but once we have more of the mod in place we can expand it to a nice readme


MateuszGie:
Checked both changes in game, no issues.

MateuszGie (approved):

PR #32, opened 2025-11-14
Rework how Ask for Money interaction works

MateuszGie:
add logic for AI to reject it if result would be actor being richer from them or if actor has heaps of money already, add prestige effect for both sides, make AI ask for it if they have loans; AI reasoning has tooltips

This is how interaction looks for country with a lot of money trying to get money from another rich country:
image

logic is moved to AI acceptance instead of just enable/disable so there is option to use it against recipients diplo reputation at cost of favors and prestige and to keep options open in the multiplayer. Checked against error.log

### DIFF
--- a/in_game/common/country_interactions/MnT_ask_for_money.txt
+++ b/in_game/common/country_interactions/MnT_ask_for_money.txt
@@ -1,0 +1,156 @@
+﻿REPLACE:ask_for_money = {
+	type = diplomacy
+
+	category = CATEGORY_ECONOMY_ACTIONS
+
+	diplomatic_cost = ask_for_money_cost
+	
+	ai_tick = daily
+	ai_tick_frequency = 360
+
+	potential = {
+	}
+
+	cooldown = { 
+		type = ask_for_money
+		years = 10
+	}
+
+	allow = {
+	}
+	
+	select_trigger = {
+		looking_for_a = country
+		target_flag = recipient
+		ai_interaction_source_list = {
+			scope:actor = {
+				every_friendly_or_high_opinion_country = {
+					add_to_list = source
+				}
+			}
+		}
+		name = "choose_country"
+		column = {
+			data = name
+		}
+		visible = {
+			not = {	#Military Orders have their own diplo action with a more flavorful name and more money to gain
+				scope:recipient = {
+					giving_scripted_relation = {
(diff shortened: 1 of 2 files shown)


in_game/common/country_interactions/MnT_ask_for_money.txt
```diff
@@ -0,0 +1,156 @@
...
+					giving_scripted_relation = {
+						type = relation_type:military_sponsorship
+						target = scope:actor
+					}
+				}
+			}
+		}
+        enabled = { 
+			gold >= {  #this action cannot put recipient in debt
+			 	value = scope:actor.country_tax_base #scope:actor.country_tax_base x 4 is one possible value being transferred
```
CorlisFataneya, line 48: Would it be possible to scale the requested amount of money with the diplomatic reputation of the requesting nation, amplified by the diplomatic reputation of the paying nation. 
Therefore very reputable countries help each other more, while bellicose countries will be able to exploit this less. As this is a friendly action and not a bribe/blackmail

CorlisFataneya (approved):
You have my general approval, although I couldnt playtest it. Lack of time. 
Please check my comment in case you like it and want to modify it :)

in_game/common/country_interactions/MnT_ask_for_money.txt
MateuszGie: I guess it could be done, yes, it would be a bit opaque though.
So now what we have is max (income of paying; 4 x income of requesting), you would like to take this value and multiply it twice, with sth like (4+diplo rep of paying)/10 and then (4+diplo rep of requesting)/10? So with diplo rep of 6 effect is kept and then it changes if diplo rep is different? 

This seems a bit complicated, I would rather propose Belligerent/Consilliatory value having impact on acceptance from recipient, it seems easier and has a direct communication to a player.

CorlisFataneya: thats a good approach :)

MichaHelder (approved):
Your changes make sense to me, script looks good. 

Also tested, no errors on startup and neither after some years of observing.

MateuszGie:
as written in "help with git" channel, I cannot fined an option to merge those branches. Please click it instead and maybe let me know where it was/why I was not getting the button :)

Issue #20, opened 2025-11-11
cultural assimilation


VineFynn:
Related document:
https://docs.google.com/document/d/1JLNnX5r2243NviCgq0PhK0YSIIlmttLPUuJfoKBM9LI/edit?usp=sharing

There was a discussion on the discord about a desireable endstate for cultural assimilation.

Key points:
- [x] 1. Cabinet action needs to be slower
- [ ] 1a. could have low base amount but scale up slightly for areas with higher populations, so you arent never converting areas in china but instantly converting areas in europe
- [x] 2. vassals should not speed up the process 
- [x] 2a. Idea: same-culture vassals can't use promote culture, but overlord can use that cabinet action on the vassals territory
- [ ] 2b. Idea: No or very limited cabinet action for all countries, mostly based on national modifiers such as cultural tradition and influence
- [ ] 3. less influence of control, but not none
- [ ] 4. opinion of the culture of the assimilating (or just owner?) culture should matter

Related points:

- [ ] enforce culture should not be an option for vassal interaction (ed. possibly fiefdoms?)
- [x] we can make certain buildings add pop assimilation
- [ ] art needs to matter more

CorlisFataneya:
I would add, culture conversion should be blocked or massively hampered by Unrest.

VineFynn:
> I would add, culture conversion should be blocked or massively hampered by Unrest.

Bad news, no unrest static modifier :(

VineFynn:
Also, I can't see unrest indicated anywhere

Malorn-Deslor:
I moved the cabinet action value up to 40% rather than 10% of vanilla, was too slow to actually do anything. Control acts as a good limiter in this case as well, but has the same problems of being overly effective in some countries.

The vassal problem is certainly still a problem, still unsure on how to fix it, but it's a niche player exploit.

VineFynn:
> I moved the cabinet action value up to 40% rather than 10% of vanilla, was too slow to actually do anything. Control acts as a good limiter in this case as well, but has the same problems of being overly effective in some countries.

The problem is there's sweet fa we can do to reduce assimilation speed in sparaely populated areas by anything other than a linear amount proportionate to population- the static modifiers simply won't accept script values and neither will location ranks.

MateuszGie:
> Also, I can't see unrest indicated anywhere

[...]

PR #35, opened 2025-11-15
Assimilation speed changes

VineFynn:
Relates to https://github.com/MEIOU-and-Taxes/MnT-EU5/issues/20

Not a complete fix, since certain modifiers are unavailable and certain other changes have yet to be made, but should resolve the total homogenisation of europe by 1400


Dev2150 (approved):
I agree with the changes & they all work

Issue #27, opened 2025-11-13
balancing war exhaustion


VineFynn:
Currently we see very low values of war exhaustion even for countries experiencing spectacular defeat after spectacular defeat- we should probably address that, and evaluate how much war exhaustion effects ability to wage war

VineFynn:
I'm still not sure this is right but I have managed to get a country up to 6! war exhaustion, which is pretty bad. Maybe the numbers matter more later in the game

VineFynn:
Reduce base reduction speed

PR #36, opened 2025-11-15
Triple war exhaustion from battles, to get feedback from others

VineFynn:
War exhaustion levels are currently too low- but unlike in EU5 having it max out also causes a country to capitulate. So I'm erring on the side of caution- this number may even be too high, triple seemed to deliver some incremental increases over several large battles when I was playing England, but France's level seemed stable at 1.5-2. Unfortunately I don't have enough time to playtest further so I'm leaving it to the rest to figure out if we need to tune this number down or adjust war exhaustion in some other way.

Relates to https://github.com/MEIOU-and-Taxes/MnT-EU5/issues/27


CorlisFataneya (approved):
It will always Show all lines as changed when pull the file over for the first time or? 

Well, approved :)

CorlisFataneya:
I dont think tripling is too low.

VineFynn:
> It will always Show all lines as changed when pull the file over for the first time or?
> 
> Well, approved :)

Yep, it's a new file, so every line is a change from nothing.

Dev2150:
I wonder if you can only include a single define in a group. I have asked in the EU5 coop server

Mazutaki:
> I wonder if you can only include a single define in a group. I have asked in the EU5 coop server

Yes

PR #38, opened 2025-11-15
Fix encoding mistkake I made myself earlier and stop marry_noble vanilla error spam

MichaHelder:
2 small and harmless changes. Can and should be merged soon.tm
Doesn't alter gameplay at all.


Dev2150 (approved):
Seems alright!

PR #41, opened 2025-11-17
Add shared config folder to be used by the plotter & log cleaner

Dev2150:
image


MichaHelder:
Looking good to me, but requesting review from Jarkami and Inimicus as they understand Python way better than me :)

Veheled (approved):

PR #46, opened 2025-11-18
Allow listing modified files in the repository, with respect to a patch version from modding digests

Dev2150:
I've considered making a script that returns the list of modded files that are in the changed list file from [modding digests](https://github.com/Europa-Universalis-5-Modding-Co-op/modding-digests/blob/main/1.0.5/changes_files.md) 

image
image


Dev2150:
After renaming, it now shows the defines file!
image

CorlisFataneya:
Is there an issue for this? 

What are the benefits of doing it? 
I am not sure to See why it is beneficial.

Dev2150:
@CorlisFataneya 
> What are the benefits of doing it? I am not sure to See why it is beneficial.

When a new patch arrives, you know immediately which files to check in our repo for interferences,
as opposed to reading the release notes wall of text & guessing which files to check
or comparing 2 folders of EU5...

PR #53, opened 2025-11-19
Add new naval levies for later ages, etc.

### LINKED ISSUE #28: create levies for remaining merchant transport ship tiers
And move the cog levy to be unlocked by the tech that unlocks cogs lol

VineFynn:
Add new naval transport levies for later ages- don't include uniques because that doesn't make sense

Amend filenames to conform with new naming convention

Closes https://github.com/MEIOU-and-Taxes/MnT-EU5/issues/28


Dev2150 (approved):
Seems alright
image

PR #55, opened 2025-11-19
Covert action cost rebalance and tweak to antagonism from stealing maps

MateuszGie:
* monthly cost of sow_discontent and corrupt_officials reduced to 10% of monthly income of an actor with floor and ceiling values; 
* sabotage reputation cost increased from 5 spy power to 10. 
* change placing of antagonism bomb from stealing maps, it was place that got discovered, not it is the place that was robbed

Monthly costs cannot be for now set to be based on recipient income (only one-time costs can) hence the solution with floor and ceiling


VineFynn:
Should be possible to base it on recipient's tax base as diplo_cost = scope:recipient.script_value, unless that scope isnt available in that part of the script

VineFynn:
> Should be possible to base it on recipient's tax base as diplo_cost = scope:recipient.script_value, unless that scope isnt available in that part of the script

Welp, that doesn't work, and I have no idea why.

Otherwise, this tests fine. Well done!

VineFynn (approved):
[](url)

MateuszGie:
> > Should be possible to base it on recipient's tax base as diplo_cost = scope:recipient.script_value, unless that scope isnt available in that part of the script
> 
> Welp, that doesn't work, and I have no idea why.

There is a list of things that can be used for cost calculation, with a dedicated option to use recipient tax base, but it just don't work with monthly spending. I will post it on the forums/mod co-op.

PR #56, opened 2025-11-19
Make cost of the court generate market demand

### LINKED ISSUE #49: Court maintenance generates no goods demand
@brutusyhy:
Image

A common complaint about the current tax base system and income calculation is that increasing tax base (by expanding trade) might lead to faster growth in state spending than income generation, leading to a net loss. This loss is not just a financial loss; rather, it represents a genuine leakage of money from the economy. I attempt to fix this by implementing a mechanic where the money spent on the cost of the court will generate actual market demands. This means that: 1) There's now an explanation of where the money goes when you spend it on the court, enhancing realism and immersion 2) Increasing the cost of the court will increase market demand and enhancing the profitability of industries, making it no longer a net loss for the player.

The implementation now is very simple: I added a monthly country pulse that generate a temporary demand (modeled after the demand of Royal Court building, without books, gold and marble, since they tend to have limited supply) that lasts one month. The demand size is calculated using the default price rather than current price to make calculation easier. For now, the demand is generated at the market where the capital is located.

A proposed solution is iterating through all countries, then running `every_market_present_in_country` and if the market owner is the country then add the country's `court_maintenance` to a variable in the market capital (because locations can have variables, as opposed to countries)
When done with all countries, run another event which iterates through all markets and apply the temporary demand.

brutusyhy:
Make cost of the court generate market demand. 
For each country, the total money it spent on court expense will generate extra market demand in the market in which its capital is located. 
The demand basket is modeled after the Royal Court building maintenance without Gold, Marble and Silver, since those could have limited supply.
The actual demand size is scaled based on the current market price of the demand basket.

Closes #49


Dev2150 (dismissed):
A single demand for each market, nice. 
I played as England & in 2 different months I tried with min & max CotC slider and the demand is constant. I suppose that's not intended.
(EDIT: It's necessary to reopen the market panel)

Also would you consider 
- adding localization for the demand
- appending to the 'Cost of the court' one the fact that it costs these goods, for more transparency?

image
image

image

image

MateuszGie:
I am seeing this mismatch between market temporary demand and what is added to demand on the good level:

image

It does not go away if I wait, change screens or re-run events. Value of 1.71 is what was set at the game start, second value (1.82) is changing a bit with time. @Dev2150 - second value, the one seen in the goods screen seems to be responding to my slider. The way to see it is with Portugal at game start, so you have one capital in the market.

After the display in this temporary demand is fixed, it should be good to merge. This slider tooltip could get an information about good consumption added to a market:

image

Dev2150:
image

1 8 3 6

Dev2150:
@Malorn-Deslor 

> This slider tooltip could get an information about good consumption added to a market

yes, we have investigated about appending the consumed goods to the flavour of 'Cost of the Court' concept localization, but it does not work

brutusyhy:
My experience is that such mismatch will only go away when you look at the temporary demand of another market and go back to see your own. I'm not sure if this is something we can fix, seems like a vanilla game bug. About changing the game concept, I updated the localization to include a description, but for some reason the game doesn't pick that up.

> I am seeing this mismatch between market temporary demand and what is added to demand on the good level:
> 
> image
> 
> It does not go away if I wait, change screens or re-run events. Value of 1.71 is what was set at the game start, second value (1.82) is changing a bit with time. @Dev2150 - second value, the one seen in the goods screen seems to be responding to my slider. The way to see it is with Portugal at game start, so you have one capital in the market.
> 
> After the display in this temporary demand is fixed, it should be good to merge. This slider tooltip could get an information about good consumption added to a market:
> 
> image
>

brutusyhy:
Well, apparently when you set the temporary demand to last indefinitely, this is what you will see 😂. Earlier I tried to make them last one month, but that leads to terrible UI bug where you will always see two demands there, despite only one being active.


> image
> 
> 1 8 3 6

MateuszGie:
> Well, apparently when you set the temporary demand to last indefinitely, this is what you will see 😂. Earlier I tried to make them last one month, but that leads to terrible UI bug where you will always see two demands there, despite only one being active.


Any option to not display the temporary need at all (hidden = yes for this type of demand or anything like that)?
As for game concept loc, I will give it a look at the afternoon, this would be a very strange bug :)

Dev2150:
> Well, apparently when you set the temporary demand to last indefinitely, this is what you will see 😂. Earlier I tried to make them last one month, but that leads to terrible UI bug where you will always see two demands there, despite only one being active.
> 
> > image
> > 1 8 3 6

I have assumed it is a short-lived modifier, I kept checking it every month while making slider changes

brutusyhy:
> > Well, apparently when you set the temporary demand to last indefinitely, this is what you will see 😂. Earlier I tried to make them last one month, but that leads to terrible UI bug where you will always see two demands there, despite only one being active.
> > > image
> > > 1 8 3 6
> 
> I have assumed it is a short-lived modifier, I kept checking it every month while making slider changes

So the event works by clearing the temporary demand from the last month first, and creating a new one for this month, so you are indeed seeing different demands each month. I didn't set it to naturally expire after one month since the UI will bug out and keep showing two demands at the same time.

MateuszGie:
@brutusyhy - could you try to have two rotating modifiers? If cost_court_1 active, remove cost_court_1, add cost_court_2, if cost_court_2 active, remove cost_court_2, add cost_court_1. Otherwise, they work in an identical fashion. There may be some re-fresh which makes it that remove and add in one step never resets the display and date.

brutusyhy:
> @brutusyhy - could you try to have two rotating modifiers? If cost_court_1 active, remove cost_court_1, add cost_court_2, if cost_court_2 active, remove cost_court_2, add cost_court_1. Otherwise, they work in an identical fashion. There may be some re-fresh which makes it that remove and add in one step never resets the display and date.

Good idea! I'll try that out later

Dev2150:
> If cost_court_1 active, remove cost_court_1, add cost_court_2, if cost_court_2 active, remove cost_court_2, add cost_court_1. 

yes, with if -> else_if flow, to make event firing intermittent

Dev2150:
> Any option to not display the temporary need at all

this would create a Black Box simulation, going against the vision statement...

MateuszGie:
> > Any option to not display the temporary need at all
> 
> this would create a Black Box simulation, going against the vision statement...

What I was suggesting was to hide modifier on the market level and keep values on the goods level, such operation was possible in EU4 where you could hide country modifier but when you hovered over value of e.g. coring cost reduction, it would include modifier name on the source list. 

I am not saying we should go this way, but we need to know what options there are, as displaying faulty data is a no-go. Image for reference, I mean hiding entry in temporary demands, keeping one in goods balance. If solution with rotating modifiers works there is no need for that of course.

> > image

Dev2150:
I made it work, but although I have included `hidden = yes`, the goods demand does not go away in the market panel

MateuszGie:
> I made it work, but although I have included `hidden = yes`, the goods demand does not go away in the market panel

Where are those commits? This branch and pr has only stuff from Brutus yet. I can do review on the evening, busy day at work:/
Ok, commit came in the meantime:)

brutusyhy:
> > I made it work, but although I have included `hidden = yes`, the goods demand does not go away in the market panel
> 
> Where are those commits? This branch and pr has only stuff from Brutus yet. I can do review on the evening, busy day at work:/ Ok, commit came in the meantime:)

I just used the two temporary demands approach, and looks like it now gets correctly updated every month. I guess one thing left is to do some logging analysis and see how this increased demand affect the economic balance, especially for large countries like Yuan

main_menu/localization/english/MnT_game_concepts_l_english.yml
```diff
@@ -0,0 +1,2 @@
+l_english:
+   game_concept_court_maintenance_desc: "The court is the extended household of the [ruler|e], including all those who regularly attend on them. In the largest courts, this number can grow to many thousands, comprised of camarilla, household staff, clergy, gentry, bodyguards, harem, concubines, eunuchs, and many more roles. All these people do not come for free, and this figure represents the cost of having them around. In MEIOU & Taxes, The money spent on the court translates to increased demand (Furniture, Tools, Glass and Fine Cloth) at the market where the capital is located."
```
MateuszGie, line 2: I have tested version that @brutusyhy provided and I am happy to see that solution with rotating modifiers did work. For me, this is ready to be merged, works as designed and design is very solid. 

1) For missing loc - issue is UTF-8 BOM, after re-saving with UTF-8 BOM I have peoper loc displayed, I had vanilla string here after loading a branch:
image

2) In the future we may want to look at balancing or changing basket of goods, but I don't think we should be waiting with merge on that, goods are in flux anyway and this solution is self-balancing as it starts from money and backcalculates goods that cost that much.

3) For hidden=yes, we probably should make a report to the PDX, it may be useful in the future

4) For a solution with rotating modifiers being needed to update the loc - same thing. Both will be probably low on priority list, but it is still good to have.

MateuszGie (commented):
ready to merge after fix to loc encoding

main_menu/localization/english/MnT_game_concepts_l_english.yml
brutusyhy: Great! I have fixed the BOM issue and the game concepts localization works now

MateuszGie (approved):
All good now

PR #63, opened 2025-11-21
Remove complaints in log files about the logger events' lack of description & title

### LINKED ISSUE #61: Logger events in census file require orphan flag
If an event is orphaned (not referenced elsewhere) it requires the orphan flag to be set to yes to suppress warnings related to its lack of parentage. The chart events currently don't have that set.

Dev2150:
Closes #61 

Adds `orphan = yes` to all logger events


VineFynn (approved):
Changes as requested

PR #67, opened 2025-11-22
Clean the error log: Make called events not orphans & give titles to untitled events

### LINKED ISSUE #61: Logger events in census file require orphan flag

Dev2150:
Resolves #61


MateuszGie (approved):
Checked with new game, clean log

PR #70, opened 2025-11-23
Made culture/religion conversion rate scale with population.

Malorn-Deslor:
Basically, instead of having a base conversion rate, this now scales the base conversion rate to the total population.

This will mean smaller pop provinces do not convert/assimilate at much faster rates then large pop provinces. Base rate would require 400+ years to fully change a province culturally, with no penalties, at 100% control. Religion would require 200+ years.

New math means anything with 50% control or lower and a rural settlement will not see any change to culture at all, all else being equal.

This should make culture conversion less overwhelming to low pop areas.

Cabinet action base amount lowered, but now applies a slight modifier as well as base change.


MateuszGie:
With culture, I am feeling okay, but religious conversion changes led to complete conversion of European Jews in 40 years, I can't find them anywhere on the map. Christan sects like Waldesians, Cathars or Paulicians are between 4 and 5k total. 

Also:
image
Numbers are not adding up in the tooltip, so it is tough to say. 83 x 1.7 = 141, here we have eiher 83 x 2.7 or 123 x 1.7, either a base or a multiplier are off; alternatively overall number is not supposed to be a sum presented...

Malorn-Deslor:
There is no way to magically prevent jews from being converted under the current model without removing all conversion. Paradox needs to model some sort of 'religious resistance'. We cannot use conversion of jews as a baseline for tuning.

As to the tooltip not adding up, I have no idea, but it isn't from this change. My guess is there is something wonky with the cultural intermixing modifier. Have you factored in population growth in that area?

Could you possibly look at somewhere far less unique and see if the numbers make sense elsewhere?

Malorn-Deslor:
Looked into it more detail, this is a vanilla thing, it just not obvious when you have tiny numbers. There is some other factor influencing culture conversion. I suspect there is a part of the tooltip missing.

MateuszGie:
> There is no way to magically prevent jews from being converted under the current model without removing all conversion. Paradox needs to model some sort of 'religious resistance'. We cannot use conversion of jews as a baseline for tuning.

I think you got very focused on them, problem is not affecting any specific minority - almost all of them are gone. I've mentioned Christan sects, they are gone as well. I think that religious unity should affect conversion with negative sign (as it is additive), so closer you are to homogenous location, slower the conversion is. To keep overall numbers similar we can limit impact of control (e.g. split current impact that control has by half). In HRE, Italy etc. control is not an issue, so the conversion is very quick.

> As to the tooltip not adding up, I have no idea, but it isn't from this change. My guess is there is something wonky with the cultural intermixing modifier. Have you factored in population growth in that area?

I don't know how to factor the population growth really and I don't know what the value over the tooltip means really. I assumed it should be the same as base value with modified, not to be any post-factum metric as it changes when you apply new modifiers, so no reason to count pop growth. 

> Could you possibly look at somewhere far less unique and see if the numbers make sense elsewhere?

Yup, 36 base, +40% modifier, value correspodning to +117% modifier (78/36 = 2.17)
image

Malorn-Deslor:
Final adjustments are in, this should be ready for merging at this point.

Veheled:
> > There is no way to magically prevent jews from being converted under the current model without removing all conversion. Paradox needs to model some sort of 'religious resistance'. We cannot use conversion of jews as a baseline for tuning.
> 
> 
> 
> I think you got very focused on them, problem is not affecting any specific minority - almost all of them are gone. I've mentioned Christan sects, they are gone as well. I think that religious unity should affect conversion with negative sign (as it is additive), so closer you are to homogenous location, slower the conversion is. To keep overall numbers similar we can limit impact of control (e.g. split current impact that control has by half). In HRE, Italy etc. control is not an issue, so the conversion is very quick.
> 
> 
> 
> > As to the tooltip not adding up, I have no idea, but it isn't from this change. My guess is there is something wonky with the cultural intermixing modifier. Have you factored in population growth in that area?
> 
> 
> 
> I don't know how to factor the population growth really and I don't know what the value over the tooltip means really. I assumed it should be the same as base value with modified, not to be any post-factum metric as it changes when you apply new modifiers, so no reason to count pop growth. 
> 
> 
> 
> > Could you possibly look at somewhere far less unique and see if the numbers make sense elsewhere?
> 
> 
> 
> Yup, 36 base, +40% modifier, value correspodning to +117% modifier (78/36 = 2.17)
> 
> image
> 
> 
> 
> 
> 
> 
> 
> 

This is really a vanilla problem that is increased by these changes. We should check if it is somehow possible to give a specific pop conversion resistance but I don't think there is a modifier for it sadly.

Malorn-Deslor:
> This is really a vanilla problem that is increased by these changes. We should check if it is somehow possible to give a specific pop conversion resistance but I don't think there is a modifier for it sadly.

It isn't, currently I don't even know if conversion pays attention to who it is converting at all. I suspect paradox will add some sort of thing similar to cultural tradition vs influence at some point.

Veheled (approved):
Code and changes look good, balance probably needs more extensive playtesting and perhaps additional tweaks. Did a short observer and generally it looks good.
We should check if we can somehow give assimilation resistance to some specific conditions or culture, not sure we have modifiers for it.

VineFynn:
More playtesting can occur in develop.

PR #71, opened 2025-11-23
Changes to satisfaction from religion mismatch, dhimmi privs and orthodox religious opinions

MateuszGie:
This is not meant to close the branch, just move changes already made to develop so it is easier to work on balance.
Changes originate from verifying Ottoman situation, but are not limited to it and will affect other places.

Review should look at impact on pop satisfaction in places where there is significant cultural and religious minority (Turkey, Spain, India).

* it is done so Dhimmi privileges, if granted all can bring tolerance of heathens to zero, at a cost of tolerance of true faith and having privileges active
* Modify Orthodox religious opinion on Catholics, Sunni and Shia - discussed with Sun_Wu and LuuBluum, it makes starting opinion of Orthodox on Catholics and Islam the same (slightly negative) as closer to historical situation at game start. It affects satisfaction of population strongly.
* Defines update to reduce satisfaction malus from religion - vanilla effect was an overkill, keeping satisfaction of pops from rival religion below revolt threshold, even with all done in the tolerance direction. This effect should be now lower, still significant, but lower, about 15-20% of satisfaction. Described in more detail in the commit note.

*Fix to Theodosian Walls being dismantled from peace deal, leading to early fall of the city (reported to PDX on forum as well)


in_game/common/government_reforms/MnT_country_specific.txt
LuuBluum, line 1: How does this differ from vanilla? Needs inline comments for changes like the other files.

LuuBluum (changes requested):
Looks like it just needs some more detail in the Osman file for what changed from vanilla. Rest looks fine to me.

in_game/common/government_reforms/MnT_country_specific.txt
MateuszGie: just adds 		
ruler_must_be_commander_during_war = yes
, but you are right, I will add comments like in other files.

LuuBluum (approved):
Looks good to me now.

PR #75, opened 2025-11-25
Make diplomatic spending transfer money to nobles and generate market demand

### LINKED ISSUE #65: Income lost to lack of control not shown
Currently there is a secret income gained from lack of control

I believe this can be solved when 
- control can be changed so that it does not affect tax base anymore 
- estates can have variables

Image

### LINKED ISSUE #74: Diplomatic spending should generate good demand and transfer money to the nobility
Similar to Cost of the Court, in vanilla game the money spent on diplomacy simply vanishes without a satisfying explanation. My plan is to use a similar strategy: make the spending generate corresponding goods demand. 

However, as @MateuszGie pointed out, the spending on this slider can be high (since there are fewer ways to reduce the percent spent on it), so it might be better to split it into two parts: the first part goes to the nobility estate, and the second part generates market demand similar to cost of the court.

brutusyhy:
Similar to #56, except that some proportion of the money spent on diplomatic maintenance will be transferred to the nobles. For now, that's 50%. We can discuss the exact proportion and whether this money should be transferred to other estates as well.

As for the generated demand, I start with a simple basket of paper and jewelry modeled after the vanilla `Embassy` building, but we can discuss whether to make the demand basket have more goods, what goods to include, and whether we should have different baskets depending on certain criteria. My thought is that a single basket kind is probably better, since having two baskets will lead to two entries in the temporary demand tooltip.

Closes #74
Resolves #65


MateuszGie:
I've checked and it works as designed, demand on good and market level refreshes properly, good stuff.

1) For money transfer to Nobles, we should think about some way to display it. In this tooltip there is no place for "other sources of monthly money transfer", but this is the first one and there will be more to follow:
image
, so we should look for solution in advance.

2) Two basket goods: I don't think we will have two entries in the temporary demand tooltip if using if/else approach. Imagine that one basket uses paper and jewelery only and other uses paper, jewelery, luxury cloth, liquor, wine and gems if diplomatic spending is huge or country has specific rank (do not matter for the example). If condition is met, you apply temporary demand in its luxurious variant, if not - in basic variant. Those are just two separate temporary demand modifiers, but only one is active at the time. I think this is something to pursue as we can have better baskets of goods spreading cost of big slider across larger set of goods while not creating ahistorical results (e.g. liquor/wine in cost of court for markets with those goods banned). 

3) Impact on markets: in 1.0.7. at least, countries keep that slider low, so impact on prices was negligible. I've tried to set it to maximum in places dominated by one country and impact was still limited. I want to give it a look in later game though, as slider "price" may grow quicker than supply of specific good (which is why I was thinking of bigger baskets).

Veheled (approved):

Veheled:
Code looks good in general. I would prbably make it not limited on nobles especially if it is not a classic noble based country. Easiest is probably distributing the money weighted by their power share.

Veheled:
You can do that here before merging or merge this (better than nothing) and then make another PR.

brutusyhy:
> I've checked and it works as designed, demand on good and market level refreshes properly, good stuff.
> 
>     1. For money transfer to Nobles, we should think about some way to display it. In this tooltip there is no place for "other sources of monthly money transfer", but this is the first one and there will be more to follow:
> 
> 
> image , so we should look for solution in advance.
> 
>     2. Two basket goods: I don't think we will have two entries in the temporary demand tooltip if using if/else approach. Imagine that one basket uses paper and jewelery only and other uses paper, jewelery, luxury cloth, liquor, wine and gems if diplomatic spending is huge or country has specific rank (do not matter for the example). If condition is met, you apply temporary demand in its luxurious variant, if not - in basic variant. Those are just two separate temporary demand modifiers, but only one is active at the time. I think this is something to pursue as we can have better baskets of goods spreading cost of big slider across larger set of goods while not creating ahistorical results (e.g. liquor/wine in cost of court for markets with those goods banned).
> 
>     3. Impact on markets: in 1.0.7. at least, countries keep that slider low, so impact on prices was negligible. I've tried to set it to maximum in places dominated by one country and impact was still limited. I want to give it a look in later game though, as slider "price" may grow quicker than supply of specific good (which is why I was thinking of bigger baskets).

I completely agree with 1). This is especially important for the current estate income fix, since it uses the same way to add money to estate. That money transfer is very huge and not very transparent to the player at the moment.

About the 2), a likely scenario is where there's one giant country and various minor countries within the same market. The larger country would undoubtedly qualify for a higher demand basket, but the smaller countries will still only qualify for lower demand basket. A potential solution is to look at the total spending by all countries in that market for choosing demand basket instead of doing that individually. I can add that to the PR later today.

I agree with the 3). In general the impact from the increased demand is still unclear in later games. My assumption is that it would gradually lose prominence, since the growth in production capacity (or the ability to increase it when needed) should likely outpace the growth in total spending, but I'm not 100% sure.

MateuszGie:
2) Ok, now I understand. Yes, in that case it would be two market modifiers, Diplomatic spending (shared by all participants, for minors 100% of their spending, for big players e.g. 60%) and Luxury Diplomatic Spending (where e.g. 40% of diplomatic spending of subset of tags would go). Do you see it as problematic from performance/clarity standpoint?

brutusyhy:
> 2. Ok, now I understand. Yes, in that case it would be two market modifiers, Diplomatic spending (shared by all participants, for minors 100% of their spending, for big players e.g. 60%) and Luxury Diplomatic Spending (where e.g. 40% of diplomatic spending of subset of tags would go). Do you see it as problematic from performance/clarity standpoint?

I see, that sounds like an interesting idea and definitely doable.

Veheled:
I'll review it tomorrow. Not at home for the evening.

Veheled (approved):

Veheled:
Reviewed. Generally no problems with the code. Some points could use a bit more code commenting, but that's not a blocker for now imo.

PR #85, opened 2025-11-30
Decline of empires removed as a disaster, parts of it worked into Time of Struggle, which is a generic decline disaster.

Malorn-Deslor:
Decline of empires was a disaster designed to punish expansive empires, it had several problems. First it cared only for the empire title, making it it irrelevant for the majority of tags. Second it was very specific and extremely hard to end.

Time of Struggle removes most of these downsides, targeting nations based on size not title, with larger and less tolerant nations much more likely to trigger the disaster. The disaster can now be reasonably ended after five years with stability restored, and no longer requires esoteric things like 'control within capital region' which is very easy for some nations and hard for others.

It also tidily solves the 'low stability has no downsides' problem by creating a real risk of disaster for low stability in large nations. The disaster itself is survivable, but could easily see vassals peeling off, separatist nations rising, and your own estates trying to murder you.

I'm working on a few other disasters that target specific fail states that the current game does not punish very well, including having huge numbers of vassals that exceed you in power, but those will have to wait.


Mazutaki:
My first thoughts, stability impact looks good. Unsure if 0.005 pop growth is the right amount for 100 stability. In my most recent games I was able to consistently hit that, and even full prosperity only grants 0.0025.

VineFynn:
I can't get the disaster to start- even at -100 stability, it just doesn't appear in my alert bars.

Malorn-Deslor:
> I can't get the disaster to start- even at -100 stability, it just doesn't appear in my alert bars.

Do you have over 100 locations? It's not meant to target small nations.

```has_any_active_disaster = no
		stability < 0
		num_locations_owned_or_owned_by_subjects_or_below >= 100
		custom_tooltip = { 
			text = has_not_had_decline_of_empire_last_25_years 
			NOT = { has_variable = had_doe_disaster_variable }
		}

Malorn-Deslor:
> My first thoughts, stability impact looks good. Unsure if 0.005 pop growth is the right amount for 100 stability. In my most recent games I was able to consistently hit that, and even full prosperity only grants 0.0025.

I guess I'm assuming 100 stability would be quite difficult to reach, the decay grows quite a lot by that point.

Malorn-Deslor:
This now has a lot of stability fixes in it.

Mazutaki (approved):
As discussed in the discord channel. My disapproval to the values has been heard and discussed. I still think the modifiers added to stability and the addition of sources to decrease stability is very nice, yet i need to document that i think that those values are overtuned a bit and should be done on broader terms. Literacy having a stabilitiy malus and increasing base decay is a start, we need to include stability effects to a bunch of stuff. But that is stuff for another branch/thread, so this gets approved. All the other disaster stuff LGTM.

Malorn-Deslor:
> As discussed in the discord channel. My disapproval to the values has been heard and discussed. I still think the modifiers added to stability and the addition of sources to decrease stability is very nice, yet i need to document that i think that those values are overtuned a bit and should be done on broader terms. Literacy having a stabilitiy malus and increasing base decay is a start, we need to include stability effects to a bunch of stuff. But that is stuff for another branch/thread, so this gets approved. All the other disaster stuff LGTM.

I agree more complex effects on stability are wise, we'll see how it moves going forward.

PR #94, opened 2025-12-02
Minting

Malorn-Deslor:
Increasing minting threshold to allow more play in the minting slider. Also corrects estate income to be more accurate.


Mazutaki (approved):
LGTM test (real approval has been granted by @Malorn-Deslor)

PR #98, opened 2025-12-02
Colombian exchange - fixes to the situation

MateuszGie:
Current implementation of the situation is buggy, leading to very little usage as there are requirements about goods demand exceeding presence in the market, which do not work with updated code for demand of absent goods. Also, some problems with list triggers requesting 1M available pops. 

Additionally, this commit removes prestige penalty price for changing RGO which is unclear in what it represents and do not scale with larger nations at all (also - AI was not checking for its prestige in its logic). 

Includes changelog entries.


MateuszGie:
First pass on triggers done, but I am taking it back to draft as "unlocking" the situation only showed that it is broken underneath.

AI logic on will_do has to be re-written - as this is is a button in the situation screen which opens a nice selector, items in the selector (goods -> regions -> location) are picked when first of them clears the AI will do threshold. This means that either they never do (e.g. potatoes or wheat) or they are used in excess (tobacco) when first in the trigger list. 

Secondary problem is with what is allowed where, with chillis taking over Europe, including Russia. I think we should be on the conservative side and while _some_ production of chilli is possible in harsher climates, I'd like to remove option to replace whole location RGO in oceanic and continental climates. I understand it removes Hungary, but we want to look into climates anyway in the future, it can be re-adjusted then.

VineFynn:
> AI logic on will_do has to be re-written - as this is is a button in the situation screen which opens a nice selector, items in the selector (goods -> regions -> location) are picked when first of them clears the AI will do threshold. This means that either they never do (e.g. potatoes or wheat) or they are used in excess (tobacco) when first in the trigger list.

So we'll have to plug the ai_will_do values into an ordered list and then only assign a positive value to the highest item in the list

MateuszGie:
I am not sure, I think we can stay with the current state and choose such ai_will_do values which will give a reasonable spread. I want to work on it soon, do a simple debug mapmode showing positive ai_will_do values and order list from most unique goods to most generic.

If this do not work, I can look to more complicated solutions.

MateuszGie:
discussion https://discordapp.com/channels/278953791586304005/1447145337490903091
there may be follow up changes to winter severity affecting feasibility of some goods, but they can be done in separate PR

in_game/common/building_types/MnT_rural_buildings.txt
```diff
@@ -714,8 +714,8 @@ settlement_building = {
 	build_time = village_build_time
 	
 	modifier = {
-		local_migration_attraction = 1.0
-		local_population_growth = 0.01
+		local_migration_attraction = 1.5 #M&T from 1.0
```
Veheled, line 717: IMO that's too high. Even the 1.0 from vanilla is already super high. Overwriting basically all other migration.

in_game/common/cabinet_actions/MnT_stabilize_country.txt
```diff
@@ -0,0 +1,13 @@
+﻿REPLACE:stabilize_country = {
+	ability = mil
+
+	country_modifier = {
+		stability_investment = 0.1 #M&T from 0.05
```
Veheled, line 5: Stability already seems quite easy to come by. Wouldn't make it easier.

in_game/common/prices/MnT_00_hardcoded.txt
```diff
@@ -0,0 +1,45 @@
+﻿REPLACE:revoke_privilege = {
```
Veheled, line 1: A lot of these balance changes have no changelog. Are they meant to be in this branch?

Veheled (changes requested):
Looks to me like this branch ahas quite a few changes outside of the core premise. Mainly balance stuff. Is that intended? The core changes itself are all fine imo.

in_game/common/prices/MnT_00_hardcoded.txt
MateuszGie: No, they are not. One is production methods that since were merged to develop, but it seems that I merged stability popgrowth branch (which is strange :/) in 334fdbdd7393d6eb26b7a6e4c26e06aaf3eff51f . My bad.

What is a good way to resolve it? Pick files that I've modified, start new branch on current state of develop and remove this one?

Veheled: If you can isolate it to specific files only related to the situation that is probably the easiest way.

PR #100, opened 2025-12-03
Remove inflation from gold and silver production

### LINKED ISSUE #99: Remove Mining Inflation
Right now you get inflation from merely mining gold and silver. This is incoherent; gold and silver in the market does not magically cause inflation. Only its minting does. If all your gold winds up in jewelry, you are not causing inflation.

LuuBluum:
Resolves #99


Malorn-Deslor (approved):
Yes, is good.

PR #102, opened 2025-12-03
Automarry event

Malorn-Deslor:
This should reduce the strain on the player having to constantly marry characters. Does not affect royal line, waits until 22 to give player to act. Biyearly, player-only.


Dev2150 (changes requested):
MnT_pulse.txt

Dev2150 (changes requested):
Add BOM. Otherwise there will be complaints in the logs

Malorn-Deslor:
> Add BOM. Otherwise there will be complaints in the logs

BOM?

Dev2150 (approved):
OK
image

Issue #33, opened 2025-11-14
Prevent arctic village PM from being used outside the arctic


Dev2150:
<img width="1199" height="558" alt="Image" src="https://github.com/user-attachments/assets/fdb78473-6e75-4423-a972-ee2492017080" />

MateuszGie:
fixed with #108 merge

PR #108, opened 2025-12-05
Production methods overhaul.

Malorn-Deslor:
Various changes to production methods, including sand to weapon and tool working.

Most importantly corrects the REPLACE bug with buildings, preventing serious economy bugs.

Also resolves issue #33 as we go.


in_game/common/building_types/MnT_town_buildings.txt
```diff
@@ -41,4 +41,40 @@
...
+
+	unique_production_methods = {
+		granary_maintenance = {
+			masonry = 0.05 #M&T from 0.2
+			category = building_maintenance
+		}
+	}
+
+
+	construction_demand = granary_construction
```
MateuszGie, line 72: I thought we will reduce the maintenance in the way it was done, but also increase the cost in game\in_game\common\goods_demand\building_construction_costs.txt

current entry:
```
granary_construction = {
	lumber = 0.5
	masonry = 0.5
	category = building_construction
}
```

I am fine with having it as-is, but doubling the construction usage of masonry also looks good.

in_game/common/building_types/MnT_common_buildings.txt
```diff
@@ -1,4 +1,40 @@
...
+	category = basic_industry_category
+	employment_size = advanced_peasant_building_employment
+	
+	rural_settlement = yes
+	town = yes
+	city = yes
+
+	build_time = guild_build_time
+	
+	unique_production_methods = {
```
MateuszGie, line 16: Both methods are bringing profitability warnings:

The stone_bricks production_method has a too high base profit 33.33%, should be no more than 30.00%
0.4 x 1 (stone) + 0.1 x 0.5 (sand) = 0.45 into 0.6x1 (masonry)

The clay_bricks production_method has a too high base profit 33.33%, should be no more than 30.00%
0.8 x 5 (clay) + 0.1 x 0.5 (sand) = 0.45 into 0.6x1 (masonry)

we can reduce output to 0.5 masonry for both

in_game/common/building_types/MnT_common_buildings.txt
```diff
@@ -1,4 +1,40 @@
...
+			output = 0.6
+			category = guild_input
+		}
+
+	}
+
+	construction_demand = basic_construction_needs
+}
+
+lumber_mill = {
```
MateuszGie, line 41: The lumber_mill_maintenance production_method has a too high base profit 66.67%, should be no more than 30.00%
0.3 x 3 (tools) = 0.9 -> 1x1.5 (wood), this warning looks reasonable

The notool_mill_maintenance production_method has a too high base profit 4294867.50%, should be no more than 30.00%
no input so zero cost - debug_profit can get rid of the 2nd warning according to VineFynn

in_game/common/building_types/MnT_common_buildings.txt
Malorn-Deslor: > Both methods are bringing profitability warnings:
> 
> The stone_bricks production_method has a too high base profit 33.33%, should be no more than 30.00% 0.4 x 1 (stone) + 0.1 x 0.5 (sand) = 0.45 into 0.6x1 (masonry)
> 
> The clay_bricks production_method has a too high base profit 33.33%, should be no more than 30.00% 0.8 x 5 (clay) + 0.1 x 0.5 (sand) = 0.45 into 0.6x1 (masonry)
> 
> we can reduce output to 0.5 masonry for both

I don't really care what it thinks is a good amount of profit, if I'm being honest. It's concept is often wildly illogical and many vanilla buildings have profit much more than these thresholds. I don't plan to tweak based on what makes the debug system angry.

in_game/common/building_types/MnT_common_buildings.txt
Malorn-Deslor: > The lumber_mill_maintenance production_method has a too high base profit 66.67%, should be no more than 30.00% 0.3 x 3 (tools) = 0.9 -> 1x1.5 (wood), this warning looks reasonable
> 
> The notool_mill_maintenance production_method has a too high base profit 4294867.50%, should be no more than 30.00% no input so zero cost - debug_profit can get rid of the 2nd warning according to VineFynn

The thing is, what is 'too high base profit'? Lumber mills are not terribly profitable, before or after, and have high employment. Many vanilla buildings have base profit up in the 60% easily. I don't think anything is broken when you spend the same amount for a lumber mill that produces at best 0.4 profit (with a lumber shortage), as opposed to paying the same for a spinning guild which could easily net you 2 profit while emplying 200 men.

As to notool, it will always be infinite profit, since it has no input, but it will never be chosen since it produces very little actual money.

in_game/common/building_types/MnT_town_buildings.txt
Malorn-Deslor: Are granaries good enough to really justify them costing more? They don't seem that valuable to me, especially with food stockpile no longer boosting growth.

MateuszGie: Probably they are not, I was just thinking about their buidling to maintenance cost ratio, I think it is a piece of infrastructure that is very heavy on the construction side and light on maintenance. Probably not a crucial thing to worry about though :)

MateuszGie (dismissed):

Veheled:
Question of the unknowing. Why do we no longer use REPLACE: ? What is the background of this.

Dev2150:
> Question of the unknowing. Why do we no longer use REPLACE: ? What is the background of this.

Because since they were added, we have all been playing with broken PMs
Simply removing them makes them work again, albeit with harmless entries in the error logs

Documentation/Change log.md
```diff
@@ -58,10 +58,12 @@ Date: 2025-11-30
     - Proximity Cost of going downstream along a River: 10 → 7  
     - Proximity Cost on Frozen Water: 20 → 15  
   - Roads  
-    - Build cost scaled & maintenance market demands scaled & rebalanced  
-      - Paved roads: \+20%  
-      - Modern roads: X3  
-      - Railroads: X10  
+    - Build cost scaled & maintenance market demands scaled
```
Veheled, line 64: @Malorn-Deslor Here the changelog is missing the actual PM changes and is showing the road changes. Is the road changes in this branch? Because I dont see an associated change here.

Veheled (changes requested):
Changelog is incorrect and needs updates.

Documentation/Change log.md
Malorn-Deslor: Road changes are not in this branch.

Malorn-Deslor: There shouldn't even BE a changelog in this branch that isn't inherited from develop.

Veheled: Well most of it is inherited from develop besides the changes to roads inside the changelog. It seems a bit messed up at the moment :/

Dev2150: Yes, I forgot to update the change log regarding the road revamp before merging and I was suggested to do it here. Now that we have the GitHub action, it would have been a good reminder

Veheled (approved):

PR #110, opened 2025-12-06
Make penalties from lack of rivals more sensible

Jarkami:


Malorn-Deslor (approved):
Seems solid, I agree with going in slowly on this, we can always increase later.

PR #115, opened 2025-12-07
feat(politics): Make estate satisfaction bonus scale with estate power

### LINKED ISSUE #109: Scale bonus from estate satisfaction by estate power
Right now the bonus from estate satisfaction is unaffected by estate power. You can have a highly secular country but a happy clergy can still provide a flat 25% research point bonus. This becomes particularly egregious when you consider the vanilla law "Secular Education" under "Education of the Masses": it will always result in lower research point (10% from 10 max literacy and -12.5% from -25% clergy satisfaction) regardless of how powerful or not the clergy is. This makes little sense and results in uninteresting game mechanic where I just always try to max out their satisfaction under all circumstances.

I suggest instead make the bonus (and malus) from estate satisfaction scale with their actual power, which I guess can be achieved using monthly pulses. Hopefully the UI for estate satisfaction bonus can be modded so that we can show the scaled value as well.

brutusyhy:
Resolves #109 

Disable vanilla flat bonus from estate satisfaction, and replace it with a more dynamic bonus scaled with estate power, calculated using monthly pulses. To keep the size of the modifier more in line with the vanilla, I have scaled the base bonus to 4 times, meaning that the bonus would be the same as in vanilla if an estate has 25% of power.

Note that since I can't replace only the satisfaction block of estates definition, I had to replace the entire definitions, but the only changes are the commenting of satisfaction bonus. This should be fine since we probably will replace them anyways.

The thing left to do is to change the GUI for estate satisfaction tooltip (p1) to show the new modifier. I struggled to make it work, so I hope others can give it a try and see what can be done here. The hope is to make the panel show the list of effects of a static country modifier (p2) instead of the built-in satisfaction effect.

image

image


brutusyhy:
Now GUI displays the correct modifier, although the implementation is not optimal, and we probably need a tooltip explaining how power scales the satisfaction modifier：
image
image

Veheled:
Any work remaining on this because it is still draft?

brutusyhy:
> Any work remaining on this because it is still draft?

Sorry I was too busy with my work at the moment. I was planning to add a tooltip to explain the impact of estate power on estate satisfaction before merging this. I'll see if I can do it later today.

Veheled:
No worries I was just curious, if perhaps the status was forgotten to be change to open.

MateuszGie:
> No worries I was just curious, if perhaps the status was forgotten to be change to open.

there is also a question if country modifiers can be hidden - you can see that we have 4-5 country modifiers that are always there. Other from that, I tested the feature and it works as designed, solving gamey bonuses from 1% strong estates.

Veheled:
Modifiers can be hidden using:

hide_effects = yes can be used to make this modifier hidden.

brutusyhy:
> Modifiers can be hidden using:
> 
> hide_effects = yes can be used to make this modifier hidden.

Apparently this can only be used on auto-modifiers. I guess I'll need to manually hide them in the GUI...

brutusyhy:
image
I added a tooltip explaining the estate satisfaction impact. I'm not sure if there's an elegant way to hide the static_modifiers, however...

in_game/events/MnT_estate_satisfaction_fix.txt
```diff
@@ -0,0 +1,157 @@
+namespace = estate_satisfaction_fix
+
+estate_satisfaction_fix.1 = {
```
Veheled: I think the namespace name is suboptimal. We should standardize the naming schema somehow. For now ok but something for late.

in_game/events/MnT_estate_satisfaction_fix.txt
```diff
@@ -0,0 +1,157 @@
...
+    type = country_event
+    title = estate_satisfaction_fix.1.title
+    hidden = yes
+
+    immediate = {
+        if = {
+            limit = {
+                country_has_estate = estate_type:nobles_estate
+            }
+            set_local_variable = {
```
Veheled, line 13: Do we really need to keep those variables around ? Isn't enough to dynamically calc them via script values?

main_menu/localization/english/MnT_static_modifiers_l_english.yml
```diff
@@ -0,0 +1,8 @@
+﻿l_english:
+ STATIC_MODIFIER_NAME_nobles_estate_satisfaction: "Nobles Estate Satisfaction"
```
Veheled, line 2: Perhaps also a modifier desc would come in handy. Not sure.

main_menu/localization/english/MnT_interfaces_l_english.yml
```diff
@@ -48,4 +48,5 @@
  FOOD_STOCKPILE_ABOVE_20_UNIT_PER_POP: "[food_stockpile|e] >20 units per 1k pop"
  FOOD_STOCKPILE_AT_0: "No [food_stockpile|e] in the Market"
  mapmode_market_food_stockpile_name: "Market Food Stockpile"
- MAPMODE_MARKET_FOOD_STOCKPILE: "#T $mapmode_market_access_name$#!\n$mapmode_common_start$ the per-population [food_stockpile|e] in their market."
\ No newline at end of file
+ MAPMODE_MARKET_FOOD_STOCKPILE: "#T $mapmode_market_access_name$#!\n$mapmode_common_start$ the per-population [food_stockpile|e] in their market."
+ CURRENT_SATIS_IMPACT: "In [meiou_and_taxes|e], the impact from estate satisfaction is determined by the level of [estate_satisfaction|e] and [estate_power|e] simultaneously. \n\nThis means that a stronger estate will provide larger bonus when they are happy, and larger malus when they are not. In contrast, weaker estates might not provide any meaningful modifiers no matter how happy/unhappy they are.\n\nThe current satisfaction level provides the following impact:"
```
Veheled: Please use full telling names. Here: CURRENT_ESTATE_SATISFACTION_IMPACT

Veheled (changes requested):

main_menu/localization/english/MnT_static_modifiers_l_english.yml
brutusyhy: I tried adding a desc to the static modifier like this:
`            add_country_modifier = {
                modifier = cossacks_estate_satisfaction
                size = {
                    value = local_var:cossacks_estate_satisfaction_level
                    multiply = "this.estate_power(estate_type:cossacks_estate)"
                }
                desc = CURRENT_ESTATE_SATISFACTION_MODIFIER
                months = -1
                mode = replace
            }`
But the modifier still doesn't have any description.

PR #118, opened 2025-12-08
chore: Add autosave file mover, to store autosaves

mak8427:
For future analysis this script can be placed in the savegames folder and it will  monitor the current directory for new Europa Universalis IV autosave files (.eu5).

When a new autosave is detected, it reads the in-game date from the file,renames the file to include the date,and moves it to a dedicated "saves_watcher" directory.

It also manages the autosaves by retaining only those that are spaced at least FREQUENCY_MONTHS apart, deleting closer ones to save space

### DIFF
--- a/tools/autosave_storer/save_game_mover.py
+++ b/tools/autosave_storer/save_game_mover.py
@@ -5,11 +5,21 @@ renames the file to include the date,and moves it to a dedicated "saves_watcher"
 It also manages the autosaves by retaining only those that are spaced
 at least FREQUENCY_MONTHS apart, deleting closer ones to save space.
 """
-
-
-import os
+import configparser
 import glob
+import os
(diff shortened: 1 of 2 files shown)


mak8427:
Example:
image

Veheled (approved):

MateuszGie:
Tested, works as designed. Only thing I would change is that if you save a non-autosave game, watcher will crash, I think it should just ignore saves not complying with the naming schema

MateuszGie:
It would also make sense to have frequency (months) moved to config, so chaning this parameter, as config is covered by git ignore

mak8427:
> It would also make sense to have frequency (months) moved to config, so chaning this parameter, as config is covered by git ignore

will be solved in here: #120

PR #126, opened 2025-12-22
Location Centers of Importance

Veheled:
This PR contains the first implementation of centers of importance, without GFX / icons yet, but fully functional besides that.

Discussion about the feature can be found here: https://discord.com/channels/278953791586304005/1446944363757441087

There are 4 types of centers:
Production
Trade
Learning
Culture

Each one has 4 tiers:
Local (Area)
Regional
Continental
World

Each center represents how "good" a location is in the center type compared to all other locations of the tier / scope.


Malorn-Deslor (approved):
No obvious problems that I can find, we'll revisit balancing it after we see it in action.

PR #131, opened 2025-12-24
feat(economy): Transfer estates wealth on annexation, vassal integration and location ownership change

### LINKED ISSUE #77: Transfer estates wealth on annexation, vassal integration and location ownership change
One major leakage of estates wealth is that when a country is annexed, the estates wealth don't go to the conquering nation: it simply vanishes. To fix this, however, it would probably also make sense to make all location ownership changes entail transfer of wealth, that way there won't be an artificial difference between conquering the entire country vs conquering the entire country minus one location. 

Of course, it's impossible to make this mechanism perfect without having location level wealth tracking, but we can simply make the transfer proportional to the location's potential tax base. This would be a great step in making the economy closer to a closed loop.

brutusyhy:
Resolves #77.
On annexation, the wealth of each estate of the annexed country will be transferred to the corresponding estate of the annexing country. If that estate does not exist for some reason, the wealth goes to the treasury. I don't want to come up with some fancy logic to deal with edge cases; hopefully this will be good enough.
When taking locations in a peace treaty, a proportion of the total wealth of the losing country will be transferred in a similar way to the winning country. The proportion depends on the share of the ceded locations' potential tax base compared to the total potential tax base of the losing country before the war ends. It's a crude heuristic but I think we don't need to spend much time on this: ultimately this issue will be addressed when we have location level wealth tracking.

### DIFF
--- a/in_game/common/on_action/MnT_hardcoded.txt
+++ b/in_game/common/on_action/MnT_hardcoded.txt
@@ -1,9 +1,12 @@
 # A crude heuristic before we have proper location level wealth tracking
 # Calculate the share of potential tax base of the losing side's locations
-# Then, transfer that percentage of the losing side's estate
-# When the winning side does not have that estate, transfer it to the treasury
+# Then, transfer that percentage of the losing side's estate to the winner's corresponding ones
+# When the winning side does not have that estate, transfer it to their treasury
 # It's not very refined, just a first step to solve the economic leakage
 
+# When a peace is signed, the order of execution is
+# on_pre_losing_war => on_took_location_in_peace_treaty => on_losing_war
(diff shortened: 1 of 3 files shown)


Veheled:
Changelog missing is blocking the merge.

Veheled:
As for code quality: I would prefer not to bloat the on actions that much. Either put it in an event or scripted effect so the on actions are clean or we make multiple on action one for each purpose with their own files. I think that's not too clean though.

Dev2150:
I agree to only have references to events in the on_actions.
Also I was requested to `add some logging to see how much wealth transfer happens`. 
I believe it's better that the PR is drafted

Dev2150:
I suggest to use estate iterator & use macro in functions like `MnT_took_location_in_peace_treaty`

PR #133, opened 2025-12-28
Feature/stability popgrowth

Malorn-Deslor:
Solves a lot of little problems, and contains the full stability rework.


in_game/common/prices/MnT_00_hardcoded.txt
```diff
@@ -0,0 +1,45 @@
+﻿REPLACE:revoke_privilege = {
+	stability = 300 #M&T from 400
+	righteousness = 10
+}
+
+REPLACE:train_general = {
```
Veheled: A lot of these now cost legitimacy or army tradition etc. in vanilla. Is it intended to overwrite this change?

in_game/common/cabinet_actions/MnT_stabilize_country.txt
```diff
@@ -0,0 +1,13 @@
+﻿REPLACE:stabilize_country = {
+	ability = mil
+
+	country_modifier = {
+		stability_investment = 0.1 #M&T from 0.05
```
Veheled, line 5: IDK if making having high stab even easier is a good idea.

Malorn-Deslor: It isn't, the other stability changes make it MUCH harder to have high stability overall. This merely makes one of the ways to improve stability better to compensate for that.

in_game/common/prices/MnT_00_hardcoded.txt
Malorn-Deslor: It's not, this was done before that change, I'll wasn't aware of that, so I'll incorporate them. The downside of long design times.

Malorn-Deslor: Fixed now.

in_game/common/prices/MnT_00_hardcoded.txt
```diff
@@ -0,0 +1,37 @@
+﻿REPLACE:revoke_privilege = {
+	stability = 300 #M&T from 400
+	righteousness = 10
+}
+
+REPLACE:recruit_explorer = {
```
Veheled, line 6: A number of these changes already exist in dev, but file has a different name. Pls rename the file on this branch and align the changes.

Veheled (dismissed):

in_game/common/prices/MnT_00_hardcoded.txt
Malorn-Deslor: Yes, those changes exist on the dev, and the file has a different name because it was named wrongly, this was a correction. I will not rename the file to the prior, incorrect naming convention. That is why this commit deleted the file.

Malorn-Deslor: If you may notice, the commit name literally says 'Merged hardcoded file to proper naming'.

Veheled (approved):

PR #136, opened 2025-12-29
feat: Fix 0 control edge case with total value added calc

Veheled:
This PR modifies the control money fix such that the 0 control locations use a custom calculation to calc the profit added by a location based on buildings and RGO profits.


Malorn-Deslor:
I'm curious, what is the performance weight? Do we have any profiling?

Veheled:
> I'm curious, what is the performance weight? Do we have any profiling?

As I am still kinda unknowing on how to do profiling no. I could not feel any performance impact and the calc is spread over the AI countries via the pulse and it only affects the 0 control locations. Still if Brutus can do some profiling that would be nice.

Veheled:
> I'm curious, what is the performance weight? Do we have any profiling?

OK, did a quick profiling. For one country. It is making this event significantly more costly for nations with lots of 0 control locations. Iterating through buildings definitely not cheap. It is fluctuating but it is having a significant impact.

image
image

in_game/common/script_values/mnt_location_produced_value.txt
MichaHelder, line 1: Why is there an empty file here?

MichaHelder: As in its used in other branches afaik, this file, but creating it empty here might give merge conflicts

in_game/common/scripted_effects/MnT_income_lost_fix.txt
MichaHelder, line 1: File is in CRLF, also why the multiply with 0.95? Seems like a balancing factor but its unclear to me atm why we'd need a fixed value balancing factor. Personally I don't mind much, but leaving fixed balancing factors in the codebase seems not a good idea based on experience from v3.0.
If we need a balancing factor at all I think we'd best make something dynamic and explain it nicely to the player

in_game/common/script_values/mnt_location_produced_value.txt
Veheled: Deleted it

MichaHelder (dismissed):
There was 1 error, which I fixed now
Merging :meepo:

cheetawilliam (approved):
Looked over and briefly tested, seems to work as intended.

PR #137, opened 2025-12-30
Mepper error fixes

MichaHelder:
Mepper hate errors
Mepper fix errors

Fixes all errors on startup in develop that are not: 
- Lumbermill notools too profitable 
- Duplicated keys for buildings (apparently vanilla bug)
- variable set but not used errors (loc doesn't count according to error log but we use them in loc)


VineFynn:
Me thank mepper

Dev2150 (approved):

PR #139, opened 2026-01-05
Balance changes to Capital Location Static Modifier

Veheled:
Small balance to remove max control and population cap from capital location


MichaHelder (changes requested):
Changes look good to me but both files are in CRLF

Veheled:
> Changes look good to me but both files are in CRLF

I thought we are using CRLF?? I explicated changes it lol.

Malorn-Deslor:
Yeah, something is weird about the formatting on that file.

MichaHelder (approved):
Encoding issue fixed

Changes still look fine, results in capitals under 100% control but that seems good to me lol

Merging

PR #140, opened 2026-01-05
Major vassal nerf, making vassals require some effort.

Malorn-Deslor:
This changes vassals power calculations, causing them to correctly understand their own strength. This means holding huge vassal lands will be more challenging. It also adds a small stability cost increase per vassal, to discourage micro-vassal swarms.


Veheled (dismissed):
Reviewed, looks all fine, fixed one error log problem.

Malorn-Deslor:
Had to deal with another merge from dev on the changelog.

Veheled (approved):

PR #142, opened 2026-01-06
Food Rebalance

Malorn-Deslor:
Woe and horror to all, may famine stalk the land and economies crumble. May all who aspire fail and fall in the dustbin of history.

Food balance drastically changed...


Dev2150:
MnT_estate_buildings:76
image

Dev2150:
There must be space after the dash
image
image

Dev2150:
Here as well 
image

Dev2150:
The logic at MnT_pulse.txt:41 has to be refactored, MnT_pulse must be clean

image

Dev2150:
Misplaced effect
image

Dev2150:
image
Aren't the highlighted lines redundant?

Dev2150:
Redundancy
image

Dev2150:
`REPLACE:`?
image

Dev2150 (changes requested):
I will do in-game checks when the previously stated errors are fixed

Malorn-Deslor:
> Redundancy image

Sigh...this commit is supposed to be removing that file...I don't know why it stubbornly remains no matter how many times I try to consolidate it.

Malorn-Deslor:
> Misplaced effect image

If you would have mentioned the brackets, it would have been much easier to understand what you meant here.

Malorn-Deslor:
> image Aren't the highlighted lines redundant?

Why is it redundant? It checks to see if there is any market which meets a base qualification, ending after finding only one market which does. I was led to believe that having a preventing check in the trigger would be better than having the event execute and iterate with immediate. Was that not the case?

Malorn-Deslor:
> The logic at MnT_pulse.txt:41 has to be refactored, MnT_pulse must be clean
> image

I have no idea what 'it must be clean' means in this context. What is the problem with MnT_pulse? Also, line 41 is not in 'MnT_pulse', it's 'on_game_start_MnT'. Also this is work done by @Veheled

Malorn-Deslor:
> There must be space after the dash image image

Half the file has a space, the other half does not. Why is the space the convention in this logic?

Dev2150:
> I have no idea what 'it must be clean' means in this context. What is the problem with MnT_pulse?

I'd like MnT_pulse to only contain function calls to other `on_action`s and events, as it will have many more over time

Dev2150:
> Half the file has a space, the other half does not. Why is the space the convention in this logic?

I've shown you what happens if there is no space between the dash & text
You can use an online md viewer to check it if your IDE doesn't have a preview function

Dev2150:
> > image Aren't the highlighted lines redundant?
> 
> Why is it redundant? It checks to see if there is any market which meets a base qualification, ending after finding only one market which does. I was led to believe that having a preventing check in the trigger would be better than having the event execute and iterate with immediate. Was that not the case?

The way I see it is that if the iterator does find a favourable case, then it has to go through the iterator again, slowing the game down

Dev2150:
> If you would have mentioned the brackets, it would have been much easier to understand what you meant here.

True. If you use PyCharm (free edition) + [Paradox Language Support](https://plugins.jetbrains.com/plugin/16825-paradox-language-support), the entire `contruct_building` effect would be in red and notice it's in the trigger block

Malorn-Deslor:
> > Half the file has a space, the other half does not. Why is the space the convention in this logic?
> 
> I've shown you what happens if there is no space between the dash & text You can use an online md viewer to check it if your IDE doesn't have a preview function

Was that what the second screenshot of the crowded text was? Because you did not in any way say anything about what that was. 

I wasn't formatting for markdown, nor as far as I'm aware was anyone formatting for markdown. Also the part of the changelog you linked is not mostly from this PR.

I really think there needs to be a distinct difference between problems with a PR, and formatting and style conventions which are arbitrarily decided. I would be rather unreasonable if I went through demanding that all the formatting 'looked right' in notepad, for example. While I can preview things in MD, I saw no reason to do so automatically, since 95% of the people viewing the changelog will not be using MD.

I don't mean to be unkind here, but this is not stack overflow. Endless assumptions about formatting do not benefit our cause. If you would like to make a case for supporting MD for ease of use, I don't think I would disagree, but without that case being made, calling people out for it is a bit troublesome, especially with no explanation.

Malorn-Deslor:
> > I have no idea what 'it must be clean' means in this context. What is the problem with MnT_pulse?
> 
> I'd like MnT_pulse to only contain function calls to other `on_action`s and events, as it will have many more over time

Well, I can see your reasoning, but again that is not MnT_pulse, it's on_game_start_MnT. Now the same logic may apply to on_game_start, but then we just have links to events, requiring more cross-referencing. Again, I don't personally have an objection, but this was in no way a standard that was established or expected. Referencing it so obtusely without explanation comes off as harsh and dismissive.

Dev2150 (approved):

PR #145, opened 2026-01-08
Feature/rgo substitution

### LINKED ISSUE #143: Replace RGOs with buildings

### LINKED ISSUE #154: Make societal value (land) affect RGO buildings

Dev2150:
[Discord thread](https://discord.com/channels/278953791586304005/1451863481660870862/1451863481660870862)

#Resolves #143
#Resolves #154


Mazutaki:
Is this the pull for the issue #143 ?

Sun-Wu:
Did you test this in game? I loaded it up and it is not ready, there are problems with it that we need to figure out first. For example, previously I could build fruit orchards in a wide variety of places where I did not have fruit RGO access as an expensive way to substitute goods, Also it shows the goods almost everywhere in the Production screen

Mazutaki:
@Sun-Wu It is still in the draft phase. No indication yet that this is intended to be merge ready.

MichaHelder (dismissed):
Tested extensively, both for errors and in-game
Branch is in way better shape then develop, and should imo be merged.
It might not be perfect, but what this branch tries to achieve has been achieved. I'm not sayng things won't need tweaking here and there, but its very solid all around. 
Approving

MichaHelder (dismissed):
Re-review after 1 commit with some changes to GitHub actions file: 

Tested extensively, both for errors and in-game
Branch is in way better shape then develop, and should imo be merged.
It might not be perfect, but what this branch tries to achieve has been achieved. I'm not sayng things won't need tweaking here and there, but its very solid all around.
Approving

Sun-Wu:
Overall, I have playtested it and I heartily endorse this. I would appreciate it if we tweaked the books production, to have a production method for printing press shop to have a way to use woodblock printing at an output penalty instead of being blocked if you lack lead. I also think the labour to output ratio needs to be reviewed as going from Scriptorum to Printing Press shop actually uses more labour per output which is strange (3.3x output but 5x the labour). Perhaps the labour issue could be fixed by using the large_religious_building_employment for Scriptorum so it uses 100 labour instead (printing press shop would then 3.3x output for 2.5x labour making it more efficient). These are suggestions and in no way meant to object to merging,

Veheled (approved):

PR #146, opened 2026-01-09
Make upstream cost the same as downstream

cheetawilliam:
both market and proximity equalized. Market river discount at 0.5, could reduce to 0.7 to reduce market bordergore.


MichaHelder (dismissed):
Checked for code quality and all is good. 
I missed the discussions on this but if this is according to how the team agreed then we are good to merge. 
Approving

cheetawilliam:
reduced to 0.7 market discount and made rivers as good as tier 2, not tier 3 roads. Ready to merge

MichaHelder (approved):
All good now, merging

PR #147, opened 2026-01-11
Feature new production methods (charcoal and naval supplies)

Sun-Wu:
Added several small new features including
1) New PMs for the charcoal maker building allowing them to build on woods/jungles or forests without needing lumber
2) Added new PMs for naval supplies buildings, allowing production without tin at a reduced rate and also removed copper as a requirement for naval supplies (light and heavy ships still require copper separately). 

I tested my code and it works, however I am unsure if I did the commits and pull request right


Malorn-Deslor (approved):
I don't see any outstanding problems. Values might need to be tweaked later, but that can be done in dev.

PR #151, opened 2026-01-14
Fix/remove time of troubles

Sun-Wu:
Removes Time of Troubles because it is cursed and bad

### DIFF
--- a/Documentation/Change log.md
+++ b/Documentation/Change log.md
@@ -187,6 +187,7 @@ Date: 2025-11-30
 ##### Disasters
 
 - ‘Decline of empires’ removed, now is ‘Time of struggle’, applies to everyone with more complex logic and less impossible to escape.
+- ‘Time of Troubles’ removed, it was bad in vanilla, and needs to be fundamentally redesigned to be reimplemented
 
 ##### Warfare
 
--- a/in_game/common/disasters/MnT_time_of_troubles.txt
+++ b/in_game/common/disasters/MnT_time_of_troubles.txt
@@ -1,0 +1,182 @@
+﻿time_of_troubles = {
+
+	image = "gfx/interface/illustrations/disaster/time_of_troubles.dds"
+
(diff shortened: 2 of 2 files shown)


MichaHelder (dismissed):
Quite some content in Vanilla needs work, but that's no reason to just disable the entire thing. If something needs changing either because it wasn't good in the first place or because its needed because of M&T changes, then change it. But don't just throw out the whole thing.

Malorn-Deslor:
> Quite some content in Vanilla needs work, but that's no reason to just disable the entire thing. If something needs changing either because it wasn't good in the first place or because its needed because of M&T changes, then change it. But don't just throw out the whole thing.

This was a problem in EU4 M&T development, where lots of bad content was left in the game because remaking it would take vastly more effort than removing it. This is flavor content for Russia, representing a unique event. It is very badly and broadly implemented currently, being no more unique than a random disaster, except far harder to resolve.

Ideally it should be remade into something more effective. Do you want to spend 5-10 hours doing so? Would you find that a good use of your time? If it is, I am sure you both can work together to do that, and I'm sure that Sun has plenty of possible ideas to start doing so.

The event is still there, we can rework at any time we want. It firing is, from what I can see of it, a net negative for the game. Are there any functional problems with this PR? Will it not work in some intended way?

Malorn-Deslor (approved):
Well alright then. Since everyone agrees that fixing it is not a good use of time currently, we have our answer.

LuuBluum:
> Well alright then. Since everyone agrees that fixing it is not a good use of time currently, we have our answer.

You did see the first comment objecting to this PR, right?

Malorn-Deslor:
> > Well alright then. Since everyone agrees that fixing it is not a good use of time currently, we have our answer.
> 
> You did see the first comment objecting to this PR, right?

I did, I also made a detailed reply that asked several follow up questions. I received no answer.

PR #161, opened 2026-01-16
Add missing import without which the plotter does not work

Dev2150:
Small fix that enables the plotter to work again


Dev2150:
The last commit is not relevant to the topic, but I forgot to use `chore` instead of `fix`

Veheled (approved):

PR #177, opened 2026-02-06
Remove mass enslavement

LuuBluum:
Removes mass enslavement of all wrong-religion pops in Muslim (and other) countries.


MateuszGie:
I've checked this commit, there are no slaves in the world at the game start, except of two locations in Nubia that have 5 thousands each.

in other places, they appear on the constant rate, e.g. Ottomans are gaining 100 per month until some kind of a steady state is reached. It seems to be working as designed and solving the issue, even if we have a bit of an additional transient state at the beginning of the game.

MateuszGie:
Result from 1382:
image

MateuszGie (approved):

PR #188, opened 2026-03-13
Remove mass enslavement again (#177)

LuuBluum:
Removes mass enslavement of all wrong-religion pops in Muslim (and other) countries. This got killed on a mishandled merge.


Malorn-Deslor:
It's crude, but better a crude solution than a glaring problem.

Malorn-Deslor (approved):

PR #189, opened 2026-03-14
Block tribal pops from being promoted while filling RGO vacancies

LuuBluum:
Right now pop promotion is set to extreme levels to fill building vacancies under the presumption that pops only promote to fill the vacancies present due to RGO building creation.

This is not true; tribal pops will promote. Hence, they are presently eradicated on the first month. This change blocks their promotion.


Malorn-Deslor (approved):
Looks good, there is no reason for tribals to promote in the first month.

PR #190, opened 2026-03-17
Rosth's AI tweaks

LuuBluum:
It's a bit mangled in how we managed to arrive here (had to copy their changes locally and push them myself; they still need permissions it seems), but this is the full suite of Rosth's AI changes that they've been making.

### DIFF
--- a/Documentation/Change log.md
+++ b/Documentation/Change log.md
@@ -41,6 +41,9 @@ Date: 2026-01-20
 
 - Stability map mode works
 - Backend error fixes
+- Fixed the conquistador countries not triggering the annexation event when they conquer more than 8 locations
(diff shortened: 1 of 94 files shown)


Malorn-Deslor (changes requested):
Holy hell...ok. So I'm seeing a lot of what looks like redundant items in files, to the point where I can't even parse what is changed. That or changes are made without commenting, or both.

The entire defines file is sitting here, and almost no use of replace. This entire thing needs drastic cleaning before it can even be considered for a PR. I think this might be jumping the gun a bit.

Files should be trimmed down to just things they actually change, with those changes commented with the original values. REPLACE should be used on anything that isn't buildings.

PR #197, opened 2026-03-21
Update RGO workforce to use peasants and align GUI with food consolidation

mak8427:


mak8427:
It should be ready  i also added some missing goods into the gdp calculation

Malorn-Deslor (approved):
Looks good, merging.

PR #198, opened 2026-03-24
Books, furniture, and lacquerware were all at base price.

Malorn-Deslor:
Quick hotfix for price issues with three goods.


cheetawilliam (commented):
We somehow missed this, fixed now.

cheetawilliam (approved):
there, this is the aproval button, yes, it's fixed now

PR #200, opened 2026-03-31
fix(error): fix uni error spam because no owner

MichaHelder:
We were scoping to the owner of possible Locations to build a uni, without checking they were owned first


cheetawilliam (approved):
looks good

PR #201, opened 2026-04-01
Steal maps spy action tweaks

MichaHelder:
Old Steal Maps: dropped 10 antagonism bomb in the place the maps are about
New Steal Maps: adds opinion malus of 40 with the country you stole the maps from

Profit?
Screenshot 2026-04-01 132836


Malorn-Deslor (approved):
Simple one and done, approved.

PR #204, opened 2026-04-05
Feature/trade flow rework

cheetawilliam:
This touches on two things, pathing cost and the Indian Ocean trade specifically.

For pathing, goods weight has been lowered by 10x, but the pathing pixel cost has been increased by 40x. Rivers reducing it by half, oceans by 3/4ths. Resulting that the across ocean distance is nearly the same as in vanilla but the cost over land is much much higher. This results in trade flows actually following rivers. And this combination makes shorter range trades much cheaper than long range ones in comparison to basegame.

Aditionally the minimum trade profit for the AI to concider trading has been lowered by slightly over half to compensate for the lower good prices. And temporarily untill vanilla fixes it or we do, trade efficiency has been slashed to 1/10th to avoid any trade loops.

For the Indian Ocean vision has been extended to most tags to most locations around the Indian ocean to allow them to trade with another. Along with a small trade range buff to get them across. As with how trade works currently no trade from Indonesia can make its way through India.


MichaHelder (approved):
Playtested quite a bit, talked through it with William and checked for errors. 

All good, ready for merge. Approving but not hitting button to give others the chance to take a look

Sun-Wu:
Also playtested it, seems like an improvement and the economy seems to work better

PR #205, opened 2026-04-06
Balance/tribes pastures

LuuBluum:
Reworks tribal pasture balance


cheetawilliam (approved):
It'll improve on the current situation, edge cases need to be touched up in a different way.

PR #206, opened 2026-04-07
balance(economy): move 25% of Discovery Age Iron boost to 1 age earlier

MichaHelder:
- Splits the boost from Age of Discovery to have some iron boost in the Age of Renaissance
- Iron/Tool shortage begone (you are not fun)


cheetawilliam (approved):
should smooth out the curve a bit, good change. Will see about replacing blast and coke furnace advances later with better PMs to avoid overly high modifier stacking.

PR #207, opened 2026-04-08
Sand rework

iggynatich:
Removed sand prerequisite from most of industries, transformed sand to silica sand for glassmaking and metallurgy


iggynatich:
I've commented sand use, could remove it entirely from files

cheetawilliam (approved):
checked the code, looks good, any other kind of flux or sand changes can be wrapped up in whatever larger common recource rework we're doing. But this is an improvement for now.

PR #208, opened 2026-04-08
fix sand dependency in bog_iron_smelter

iggynatich:


cheetawilliam (approved):
fix previous oversight, it good

PR #210, opened 2026-04-08
add population glass demand from 1.2

iggynatich:


cheetawilliam (approved):
works flawlessly, might as well get ahead on the patch here.

PR #212, opened 2026-04-09
fix(error): remove unnecessary <estate>_demand entries

MichaHelder:
These lines were not necessary yet caused errors -> remove them


cheetawilliam (approved):
Looks good

PR #213, opened 2026-04-10
Balance/market attraction

cheetawilliam:
- Remove the market attraction impact from prestige, trade income vs tax base and staffed building levels to stabalize market size
- Reduced market attraction from trade attraction buildings by roughly half. Whilst increasing effects of same province/area so markets don't get too small and crowded out by others.
- also lowered the market attraction from market center status by roughly half.


mak8427:
I would like to then add credit avaiability as a source of market attraction

Jarkami (approved):
LGTM

PR #215, opened 2026-04-11
Balance/rgo tweak

cheetawilliam:
Increase the max levels of all limited RGO's, and doubly so for minable goods. 
Offsets the increase for mining by having each level cost 110% more than the previous level. (can't go lower than that because of a bug)
And moves some of the iron production bonuses to PM's so the productivity just by tech doesn't go to 300% making other buffs worthless.
Also fixed localization of the iron PM's

Limited RGO's went up to 1 base
mining to 2 base

Think it just needs a bit of testing if there's now not too much, but I'd rather increase the potential from pop/dev than make undevelopped locations worth less than they already are. 

Fully expect this all to be different once we have a full mineral/mining rework.

For things like gold and silver, with our higher minting levels there were severe shortages before blocking construction of capital buildings. That might be over now.


Malorn-Deslor (dismissed):
Ok, mostly good. A couple of things I noticed. The ideal RGO bonuses doesn't exist for buildings that don't have non-ideal RGO construction. The base amounts are multiplied by 2x, so making another 2x on top of that would be 4x. That should be removed.

We also need PM increases for the non-mining iron building, to keep pace since they no longer can gain productivity bonuses. As for the scaling prices, I like them, though if we can't get it to something more sane like 50%, reduce base price to 20 or something, Also see how it interacts with building material prices, if not multiplicative, then just isolate it entirely to make it tidy.

Great trick on the scaling price, I think once we have that tamed using it for certain things that realistically might scale like mining will be great in a lot of situations.

mak8427:
image
image
can we have this into develop ? it' hard to judge the balance of my implementation in my branch during the middle game (the pics are from the develop branch)

MichaHelder (approved):
Had a few issues/comments, communicated with William via Discord and he fixed them all. 

All good now, approved and merging :)

PR #216, opened 2026-04-12
fix(performance): REPLACE vanilla priv that waste performance like crazy

MichaHelder:
Vanilla priv was checking in potential every character. So every nation will check every character every time AI checks privs. Massive waste of performance. Move the character check of this privilege from potential to an allow section.


cheetawilliam (approved):
nice find, good work

PR #217, opened 2026-04-17
Fix peasant enfranchisement

LuuBluum:
Missed a rather important snippet in the replace there. Double-check your compatching, folks!


mikejaklitsch (approved):

MichaHelder (approved):
Good fix, approving

PR #219, opened 2026-04-19
fix(gui): add script-driven income history chart

mak8427:
## Summary
- Replace the economy history plotline with a script-driven chart pipeline and custom GUI bars so history can be fully controlled by scripted variables.
- Add 50-point historical rendering with a dedicated live bar, monthly update flow, and frequency/metric toggles backed by scripted effects and scripted GUIs.
- Initialize and refresh chart data streams robustly (including yearly aggregation and display buffers) to avoid stale or zeroed graphs.


mak8427:
image
image
image

mak8427:
Moved to #224 and #225

PR #220, opened 2026-04-27
Change modifier for tribal strongholds

LuuBluum:
Having this be a penalty to control makes the privilege so awful that the AI will gladly revoke it even if it plunges itself deep into negative stability. Additionally, it cuts against the very description of this privilege.


Malorn-Deslor (approved):
Looks good, glad you cleaned it up.

PR #221, opened 2026-04-28
Balance/ai marriage

LuuBluum:
Vanilla AI logic for non-royal marriages left a lot to be desired, including a weird snippet that effectively made the AI never consider lowborn marriages ever. Now the AI weighs marriages against life expectancy, and also accounts for whether or not they can have multiple spouses.


MichaHelder (approved):
Looks good, nice improvement, no errors. 
Approved

PR #222, opened 2026-05-03
Balance/military

LuuBluum:
Does significant refactoring of military balance, namely levies.


MichaHelder (approved):
Looked through the entire diff and all seems fine. I didn't have the time to follow discussions or playtest, but from a syntax quality side all seems in order. Also checked error log on startup which is also fine. 

Considering this is a balancing PR, syntax is all good, and balance can be easily changed by tweaking numbers later, I'm approving the PR.

PR #228, opened 2026-05-10
Fix building tooltips and diplomatic spending transfer

mikejaklitsch:
## Summary
- Restore building tooltip detail parity and revert cathedral PM to 1.1.10 state
- Fix diplomatic spending not actually transferring gold to nobles estate (estate scope prefix `root.` was blocking execution)
- Add delayed display variable for diplomatic spending tooltip with visibility gating (hide when value is 0)

### DIFF
--- a/Documentation/Change log.md
+++ b/Documentation/Change log.md
@@ -39,6 +39,7 @@ Date: 2026-01-20
 
 #### Bugfixes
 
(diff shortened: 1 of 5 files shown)


MichaHelder (approved):
Tested in-game, works and nice fix. 

Approved

PR #229, opened 2026-05-11
fix(triggers): update city and town illustration triggers to fix error spam

mak8427:


MichaHelder (approved):
Good simple fix for some Pordox error spam, approved.

PR #232, opened 2026-05-16
feat(balance): rebalance goods demand from cost of the court

MichaHelder:
- Old system demanded ridiculous amounts of just a few goods in mid to late-game
- This commit lets it demand goods for only 10% of the court slider
- The basket of goods demanded has been broadened, less per good, more different goods.


MichaHelder:
Ping @mak8427 to notify I am addressing the issue you found in your own tests :)

cheetawilliam:
The most important part was that is that it caluclates of the position of the slider, not of how many ducats are actually spend, which caused a lot of money to be created out of thin air as any positive legitimacy modifiers were lower the ducat amount paid but kept the amount of goods created the same.

At 10% that shouldn't happen anymore in 99% of cases. Only delete some money, but preferable to printing.

mak8427:
> Ping @mak8427 to notify I am addressing the issue you found in your own tests :)

Thanks!

cheetawilliam (approved):
Diplo slider needs a similair kind of treatment, but this works for cost of court.

PR #234, opened 2026-05-17
Stop randomly spawning eunuch children

MichaHelder:
As the title says

Had this happen in my game, no more: 
image


cheetawilliam (approved):
technically fine, agreed. Should be represented differently.

PR #235, opened 2026-05-20
Update to 1.2.4 + add peasant enfrachisment effect where missing

MichaHelder:
- Update enfranchisement in privileges we replaced (add it where it should be, plus remove power effects once because it said in comment replace with enfranchisment when added)
- Add vanilla annexation speed for (de)centralize value
- Update to serfdom vs free subject (add enfranchisment, organize so its somewhat readable with a few comments)
- 1.2.4 compat: update 0.7_cities and building
- 1.2.4 update location_window.gui


in_game/common/societal_values/MnT_00_default.txt
```diff
@@ -29,34 +31,33 @@
...
+		#Levy effect, bigger and faster recovery
+		levy_recovery_modifier = 0.25 #M&T from 0.5 Vanilla
+		peasants_estate_levy_size = 0.25
+		#Main effect is now enfranchisement
+		global_peasant_enfranchisment = -0.33
+
 		control_importance_modifier = 0.1
 		court_language_is_common_language_importance_modifier = -0.25
-		#Estate Power modifiers for lower classes
-		global_peasants_estate_power = -1 #added in MnT
```
LuuBluum, line 41: Do not remove these.

in_game/common/estate_privileges/MnT_peasants_estate.txt
```diff
@@ -47,7 +49,8 @@ REPLACE:peasants_represented_in_parliament = {
 	country_modifier = {
 		peasants_estate_can_participate_in_parliament = yes
 		peasants_estate_target_satisfaction = medium_privilege_target_satisfaction
-		global_peasants_estate_power = 0.25 #MnT changed from 0.5
+		global_peasants_estate_power = 0.25	#M&T from 0.5
```
LuuBluum, line 52: This needs testing to make sure their power isn't too high.

in_game/common/societal_values/MnT_00_default.txt
MichaHelder: Restored

in_game/common/estate_privileges/MnT_peasants_estate.txt
MichaHelder: Yeah that one priv I upped the power of because Vanilla upped power too. 

Pushed new commit, ignore Vanilla and keep M&T value

LuuBluum (approved):

PR #236, opened 2026-05-23
Add disease resistance reduction

Mazutaki:
Influenza and smallpox added from vanilla
Bubonic increased to value of 1.2.4


cheetawilliam (approved):
Catching up to some vanilla code aditions/changes, so is good

PR #237, opened 2026-05-24
Fix bugs/feedback from initial release

MichaHelder:
Quite a few different fixes. 

I wont be making 5+ branches to bugfix from player feedback D: 
So just put it all in one branch

Changelog as added: 
- Make every level of RGO Building give 1 free building level in its location (so you can't effectively get less available building levels from building up rural resource extraction stuff)
- Similarly for farming villages and tribal transhumant_wool_pasture buildings, give 1 free building level per level as we sort of force the player to have these buildings
- Fix Fish RGO Building not fulfilling upgrade potential if Location defined as fish RGO
- Fix cost increase per level in RGO modifier (mostly increased cost for mines per level) showing extra cost in green -> make properly red

Also: 
Fix error on startup from value for loc
Fix startup errors for loc (mainly PMs)
Reorganize RGO PM file (pain)
Start localizing RGO PMs (more pain)


Sun-Wu (dismissed):
Approved, I have playtested it and works nicely

Sun-Wu (approved):

PR #239, opened 2026-05-28
Balance/tribes

LuuBluum:
Rebalances tribal levies. Now they're 15-sized levies across the board. Introduces more tribes in the Steppe to compensate, since there are regions oddly devoid of tribal pops.


MichaHelder (approved):
Really nice solution for the lack of tribal pops in parts of steppe :)

We might want to expand the changelog a bit to explain to the player why they don't see any slaves btw :P

Overall good and welcome changes, approved

PR #240, opened 2026-05-28
rebalance attrition

cheetawilliam:
Reduces max attrition to make movement not too expensive in the high north morale wise. Adds a tiny bit of attrition in all climates to represent moving armies even without combat is taxing. Standing in ZOC of forts that you own removes attrition.

Edit: had to remove all attrition from all climates, winter still gives attrition. When suffering attrition you don't reinforce, and some tags don't have forts so they'd never be able to reinforce.


MichaHelder (approved):
Shame that this change is needed, but it is. 

Approved

PR #241, opened 2026-06-01
Feature/loc and changelog tweaks

MateuszGie:
Minor changes to tooltips for better clarity, removed changelog entries about map modes that vanilla took from the mod since

image

image


MichaHelder:
Syntax wise all good, and the extra explanations are nice. The only thing is that I feel like the flow of the sentences could be improved a bit, as in the English grammar and just how English people write. Its for sure minor nitpicking, but some sentences read like they are written by someone not native to English. 

@Jarkami for example, might you be able to take a look at the grammar/flow, as a native English speaker. Or would you consider this to be too nitpicky?

Edit: for sure it has improved from how it was btw, its just that maybe we should improve it somewhat more before merge now that we are at it?

Mazutaki:
I would add a "the" to "of _the_ maintenance cost" and also at "affects _the_ assignment_s_" and maybe make assignment plural. But thats a non native speaker perspective again :D

MichaHelder (approved):
Looking good. There were a few minor notes about Localization tweaks but those can be fixed later. Approved.

PR #242, opened 2026-06-02
Feature/urban concentration

Jarkami:
This PR needs extensive gameplay testing before we even consider merging it.

- Remove location building limit
- Changes to industry-specialization urban rights
  - Urban Rights for industry specialization now completely uncap their associated industry levels rather than giving inherent production bonuses/maluses
  - The Borough Privileges advance, which unlocks generic specialization Urban Rights, has been moved from the Age of Discovery to the Age of Renaissance (banking tree) so it is available earlier
- Change scaling production efficiency to 1% per additional building level, not modified by current age


Jarkami:
Still needed:
- Need to figure out if we want to do something with marketplace caps as well. Should they always be uncapped? Should they be uncapped with some urban right(s)? Or should we just leave them as is?
- There are a few vanilla events that trigger for the old vanilla generic specialization urban rights that this PR replaces. We should update these so they are sensible and reference the new rights (or remove them).
- Cleanup of various preexisting building slot bonuses since these are now irrelevant

MichaHelder (approved):
Already released, ready enough to be merged. Might need more tweaks later, therefore not deleting the branch for now.

PR #243, opened 2026-06-03
feat: Allow increase control cabinet if control not perfect

MichaHelder:
It was somehow limited to being under 95%
Action would also stop if going over 95% on one province minors 
Just allow now whenever not 100% OR 100% and cabinet dude busy (to allow one to keep capital at 100%)

This specifically fixes reported cases by players where they have only one province (e.g. Free City in the HRE), and because of M&T changes their control wasn't 100% by default. 
They could initiate the cabinet action to increase control, but the action would stop whenever it reached 95%, which is the Vanilla threshold. 
This PR both allows the cabinet to always continue, and also allows the action as long as control isn't perfect because it seems arbitrary to just disallows between 100 and 95%


MateuszGie:
Everything looks fine, is there no risk that AI will lock its cabinet action to this while not being a one province tag? It is not the worst usage of cabinet action, I guess but should we consider that option?

cheetawilliam (approved):
Looks good

PR #245, opened 2026-06-26
Refactor EPBM to do everything at runtime.

mikejaklitsch:
Massive architecture change that should make this far easier to maintain. Production methods of buildings are now dynamically grabbed at runtime and the relevant goods demand is priced and accumulated to a cache on a variable map on the market center, key = production method; value = raw cost of pm. Market Access and Employment are accounted for before the value is cached and then added back in per building whenever a building hits the cache. This keeps the whole system performant at scale even with the extra overhead of having to check through the pm goods.

Because of this change, all the architecture that was required to store production method data is redundant and was removed. Everything that needed to be generated, or hooked into buildings is removed. The only maintenance required going forward is maintaining the blacklist map of fort building types because there is no way to identify them at runtime.

Estate buildings are paid by the estate in full, government category buildings are paid for by the country in full, all other buildings costs are split by estate power. Inflation costs are included in the split so each estate pays for their equal share of inflated building costs.


in_game/localization/english/MnT_gui_l_english.yml
```diff
@@ -1,4 +0,0 @@
-﻿l_english:
- MNT_EPBM_ESTATE_MAINT_LABEL: " -  From [MnT_epbm_estate_building_maintenance|e]"
- MNT_INCOME_GRAPH_VIEW_TT: "Change income graph view"
```
MichaHelder, line 3: We still need these 2 income graph localization strings :P

MichaHelder (commented):
Error log all good except for those 2 missing income graph lines. 

Will try it out a bit in-game, looking good but code changes too big to effectively review by only looking at them.

in_game/localization/english/MnT_gui_l_english.yml
MichaHelder: Confirmed fixed with commit ed3413fb37dc6dcc033085e213bc9a92c099548e

Issue resolved

MichaHelder:
Confirmed savegame compatibility after 3 to 4 months of adjustment with latest commit 7b3ef3d30acee88fe6f20ad56a5f5f910bace380

MichaHelder:
image
Very nice, initializes in about 2 months fully on new save

MichaHelder (approved):
Excellent work!

Found a few issues earlier but those are resolved now.

Approving to be included in test-release :)

PR #247, opened 2026-07-06
Integrate Mazutaki's Autonomous Integration mod

MichaHelder:
Asked Mazutaki to prep his changes in a branch, I just did some error cleanup. 

Should be ready for playtesting, code-wise seems fine

Description from the Steam workshop:

> Every province, depending on various factors, integrates autonomously.
> Cabinet action has been removed, parliament issue for integration has been kept - i quite like it.
> 
> How Integration Speed is Calculated:
> Final monthly integration progress is determined by multiplying several factors together:
> 
> Base: If next to integrated/core (or in a province with 1 seazone difference)
> x Proximity
> x Cultural Opinion
> x Religious Tolerance
> x Global Integration Speed Modifier (e.g. Tech)
> x Area Integration Tech grants an additional multiplier (currently 3x)
> 
> The base beeing dependent on beeing next to a core/inegrated makes the integration "spread out" from next to already integrated provinces. BUT hardcapped at 100 years max. So after 100 Years, a province WILL be integrated, regardless off circumstances.
> 
> Capitals are always instantly and fully integrated.


Documentation/Change log.md
```diff
@@ -416,3 +416,18 @@ Date: 21/05/2026
...
+- Every province, depending on various factors, integrates autonomously.
+  - Cabinet action has been removed, parliament issue for integration has been kept
+  - Final monthly integration progress is determined by multiplying several factors together:
+    - Base: If next to integrated/core (or in a province with 1 seazone difference)
+    - Proximity
+    - Cultural Opinion
+    - Religious Tolerance
+    - Global Integration speed modifier
+    - Area Integration tech grant a 3x multiplier
+  - Integration can be slow but is hardcapped at taking 100 years
```
mikejaklitsch, line 452: CRITICAL: There are two defines which control how the AI considers expansion when holding unintegrated land. We may need to modify these if integrations are going to take significantly longer
- AI_UNINTEGRATED_LAND_EXPANSION_PENALTY = 2 # e.g. 50% unintegrated land with value 2 = -100% chance to expand
- AI_UNINTEGRATED_LAND_EXPANSION_PENALTY_MINIMUM = 5 # Needs at least 5 unintegrated locations to apply AI_UNINTEGRATED_LAND_EXPANSION_PENALTY

in_game/common/on_action/aut_on_actions.txt
```diff
@@ -0,0 +1,98 @@
+﻿@aut_version = 100 # Must match aut_effects.txt
```
mikejaklitsch, line 1: nit: This can be a script value with a static value since it's between files

in_game/common/on_action/aut_on_actions.txt
```diff
@@ -0,0 +1,98 @@
...
+
+# New game: run full initialization immediately.
+
+aut_on_game_start = {
+	effect = {
+		aut_initialize = yes
+	}
+}
+
+### Save Game Detection
```
mikejaklitsch, line 13: nit: I love when my old code smacks me in the face! I know this works so it's fine if you want to leave it, but it can be simplified. You can set a global variable to the version number with a 32 day expiration. If it doesn't exist or If it does exist, and version is outdated: initialize the system. It works the same way just a bit less code to maintain.

in_game/common/on_action/aut_on_actions.txt
```diff
@@ -0,0 +1,98 @@
...
+		every_in_global_list = {
+			variable = aut_conquered_locations
+
+			if = {
+				limit = {
+					NOT = {
+						integration_level = conquered
+					}
+				}
+				add_to_global_variable_list = { name = aut_remove_queue target = this }
```
mikejaklitsch: nit: local variable list would be better here, should self cleanup

in_game/common/scripted_effects/aut_effects.txt
```diff
@@ -0,0 +1,266 @@
...
+
+aut_apply_integration_script = {
+	owner = {
+		save_temporary_scope_as = owner_scope
+		culture = {
+			save_temporary_scope_as = culture_scope
+		}
+	}
+
+	set_variable = {
```
mikejaklitsch, line 11: nit: There are a lot of static numbers in here that could be moved to @local_named_macros so we can more easily tune things in the future or just see whats happening at a glance

in_game/common/scripted_effects/aut_effects.txt
```diff
@@ -0,0 +1,266 @@
...
+				}
+				else_if = {
+					limit = { dominant_culture ?= { cultural_view = { target = scope:culture_scope value >= negative } } }
+					add = -4
+				}
+				else = {
+					add = -6
+				}
+				
+				if = {
```
mikejaklitsch, line 104: nit: why? a comment would be nice here

in_game/common/scripted_effects/aut_effects.txt
```diff
@@ -0,0 +1,266 @@
...
+				else_if = {
+					limit = {
+						dominant_religion.group = owner.religion.group
+					}
+					multiply = scope:owner_scope.modifier:tolerance_heretic
+				}
+				else = {
+					multiply = scope:owner_scope.modifier:tolerance_heathen
+				}
+				max = 0.15
```
mikejaklitsch: Is the unbounded negative a design choice?

in_game/common/scripted_effects/aut_effects.txt
```diff
@@ -0,0 +1,266 @@
...
+
+				max = 0.15
+			}
+		}
+	}
+	
+	set_variable = {
+		name = location_integration_gain_religion
+		value = {
+			add = 1.0
```
mikejaklitsch, line 123: Just to make sure I'm understanding this correctly its
1+( ( 0.025 OR 0.05 ) * tolerance of local religion ) ?

in_game/common/scripted_effects/aut_effects.txt
```diff
@@ -0,0 +1,266 @@
...
+				max = 0.15
+			}
+		}
+	}
+
+	set_variable = {
+		name = location_integration_gain_modifiers
+		value = {
+			add = 1.0
+			add = scope:owner_scope.modifier:global_integration_speed_modifier
```
mikejaklitsch: bug?: add = [this.]modifier:local_integration_speed_modifier ?

in_game/common/scripted_effects/aut_effects.txt
```diff
@@ -0,0 +1,266 @@
...
+	#		divide = {
+	#			value = population
+	#			min = 2
+	#			max = 80
+	#		}
+	#		min = 0.25 
+	#	}
+	#}
+
+	set_variable = {
```
mikejaklitsch: Is there any reason the advance can't just get global_integration_speed_modifier = 2.0 ? Or do you want this to multiply the other integration speed modifiers as well? 
I'm a bit worried the AI won't research the advance if they can't use the cabinet action it unlocks. Have you tested this at all? If it is an issue you could add a AI weight to the advance and that would resolve it.

in_game/common/scripted_effects/aut_effects.txt
```diff
@@ -0,0 +1,266 @@
...
+			value = 100
+			subtract = integration_progress
+			divide = var:location_integration_gain
+			divide = 12
+			min = 0
+		}
+	}
+}
+
+# Add a conquered location to the global tracking list.
```
mikejaklitsch, line 215: technically this could all be tracked by a variable map from location -> integration progress but I'm not gonna make you rewrite all that for what's likely a minimal performance gain. Just pointing it out.

in_game/common/on_action/aut_on_actions.txt
```diff
@@ -0,0 +1,98 @@
...
+### Save Game Detection
+# Bouvet Island version check — runs in weather_monthly_pulse (location-scoped).
+# Bouvet Island is unownable, so its modifiers vanish when the mod is removed
+# and reappear stale when re-added. Version mismatch catches mod updates.
+# Fires once, then the version stamp prevents repeat runs.
+
+aut_init_check = {
+	trigger = {
+		NOT = {
+			has_global_variable = aut_rebuilding
```
mikejaklitsch: nit: This was a relic of when it was built in the monthly_country_pulse and needed to not run for multiple countries. You shouldn't need this variable anymore

mikejaklitsch (changes requested):
Mostly cleanup nits, but the NAI defines I mentioned are pretty critical to AI behavior and they will need adjustment if integration takes a lot longer than vanilla.

in_game/common/scripted_effects/aut_effects.txt
MichaHelder: Good point, much easier for AI to understand indeed, and needs less variables and calc on our side. Implemented with commit d165bd86097278cb2a65824d7962d6985088d9c5

in_game/common/scripted_effects/aut_effects.txt
MichaHelder: Whether design choice or not I don't care. Seems not a good idea. Capped culture at a 15% modifier and religion and a 25% modifier. Both up and down
commits 83251fd05be1f2fce832c3fb2e9c2582f382a823 and 2cb72ca12094299be14a0afb4ef298f8d98f7ec3

in_game/common/scripted_effects/aut_effects.txt
MichaHelder: Yep, with a cap (in both directions now)

in_game/common/scripted_effects/aut_effects.txt
MichaHelder: Seems bugged indeed, good catch, fixed in 83251fd05be1f2fce832c3fb2e9c2582f382a823

in_game/common/scripted_effects/aut_effects.txt
MichaHelder: Add a comment, is 'enemy' opinion

in_game/common/on_action/aut_on_actions.txt
MichaHelder: Did so in commit 840b374097de8b9fd1474eaf0acb030312a77342 but please double check, I never really used these lists before, so best someone double-checked :D

in_game/common/on_action/aut_on_actions.txt
MichaHelder: Removed the var in 213b9816cca41a14d3f194fb6e20635b9dfb7e90

in_game/common/on_action/aut_on_actions.txt
```diff
@@ -0,0 +1,93 @@
...
+
+			remove_variable = location_integration_gain
+			remove_variable = location_integration_gain_estimate
+			remove_variable = location_integration_gain_base
+			remove_variable = location_integration_gain_proximity
+			remove_variable = location_integration_gain_culture
+			remove_variable = location_integration_gain_religion
+			remove_variable = location_integration_gain_modifiers
+
+			remove_list_global_variable = { name = aut_conquered_locations target = this }
```
mikejaklitsch, line 88: not needed because it should self clean but it won't break anything

in_game/common/on_action/aut_on_actions.txt
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		clear_local_variable_list = aut_remove_queue
```
mikejaklitsch, line 91: meant to put the above comment here. This line isn't needed because this list should self clean once the effect scope is exited. But its not a functional problem so doesn't really matter

Veheled (approved):

mikejaklitsch (approved):
looks good!

PR #248, opened 2026-07-20
Feature/urban amenities

Sun-Wu:
Adds new urban amenities building to game, adds more building slots to municipalities while municipalities have had their number of slots reduced. Net this adds more slots, _but_ it comes at substantial monetary cost


MichaHelder (approved):
Looking good, already released as well. 

Might need further tweaks later but for now good to merge

PR #250, opened 2026-07-22
fix RGO zero level bug 

mak8427:
Fix various UI bugs


MichaHelder (approved):
Thanks for the quick fix, approved

PR #254, opened 2026-07-28
Make levies actually cost goods & money while raised

Veheled:
This raises the levy cost reduction compared to normal units from 1% to 75%. This will make early game war actually cost money economically. The training of levies still is free simulating that the estates pay for this...


MichaHelder (approved):
Looks good, as long as the AIs don't bankrupt themselves on their levies. But Beorsferth's initial tests suggest its fine. 

Approved

PR #256, opened 2026-07-30
Prevent nations from rejoining league war on truce

Veheled:
Prevent nations from rejoining league war when having a truce with one of the war leaders. Additionally tweaked AI weights in order to get more sensible results and not global war.


MichaHelder (approved):
Good changes, thanks for adding the comments :)

Approved

PR #257, opened 2026-08-01
Amenities and farming village tweaks

MichaHelder:
- Up Amenities required for City (4 -> 5) and Megalopolis (8 -> 10)
- Set important_for_AI flag to Amenities building_type
- Require 20k Burghers for a Location to be valid to become a Megalopolis (in addition to total pop being 200k)
- Don't let nations and estates build farming villages
  - There is a scripted system that handles that and should be the only thing handling it
 - Set version number as displayed to 0.2.6 as 0.2.5 has already been released

### DIFF
--- a/.metadata/metadata.json
+++ b/.metadata/metadata.json
@@ -1,7 +1,7 @@
 ﻿{
   "name" : "MEIOU and Taxes",
   "id" : "meiou_and_taxes",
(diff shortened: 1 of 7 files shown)


Sun-Wu (approved):
Approved, necessary changes

PR #258, opened 2026-08-02
fix(gui): add economy graph visibility toggles

mak8427:


MichaHelder (approved):
Looks good, thanks for making this so quick :)

PR #260, opened 2026-08-04
Bailiff tweak

MichaHelder:
- Re-enable Bailiffs for Urban Locations
  - Let them be a source of local Proximity (25) instead of giving maximum Control
  - Building a Local Governor will destroy any Bailiff in the Location (and make Bailiff not buildable)
  - Add this extra info to the in-game description


Malorn-Deslor (approved):
Go forth!

PR #261, opened 2026-08-04
Temp fix to ensure player can still build farming villages after blocking AI.

Malorn-Deslor:
Just as it says. The fix to prevent the AI spamming farming villages prevented players from building them. Player nations have no method of gaining farming villages at all unless able to build them, thus need to be able to build them.


MichaHelder (approved):
Approved

PR #262, opened 2026-08-08
Rework great pestilence

MichaHelder:
Needs playtesting mostly. I can fix any errors easily but for balance etc. we need testers. 

### Great Pestilence reworked
- Imaginary disease Great Pestilence is gone
- Reworked the Great Pestilence Situation:
  - Now tracks Smallpox, Influenza and Measles
  - During the Great Pestilence Situation these diseases can start spawning in the new world
  - First they can only spawn in Locations with top-overlord in the Old World, but everywhere any of the 3 diseases spread will become a valid location for its start
  - Situation-linked events are all reworked to be working off these diseases instead of a 'Great Pestilence' disease
  - Situation mapmode will display with stripes any locations that have already been affected
  - Situation otherwise functions pretty much the same


cheetawilliam (dismissed):
Probably needs tuning, but that needs more testing.

MichaHelder:
I played through it once now. The main thing seems to be that the Old World diseases don't hit the new world all that badly, plus not all that many spawn. I can force spawn 1 smallpox at the start of the Situation maybe, but otherwise game systems in how they are currently designed don't seem to result in that high dead count for the normal diseases, even though there isn't any inate resistance to e.g. smallpox

cheetawilliam (dismissed):
pappa nurgle aproves

PR #263, opened 2026-08-10
Looting will reduce dev directly, dev will decay at low enough prosperity. 

Malorn-Deslor:
As above.


cheetawilliam (approved):
Just going to need to keep an eye on the balance/results in game, but idea is good.

MichaHelder (approved):
Looks good, only worry is low dev areas not being able to really escape low-dev even more (aka not-Europe). Will probably need to tweak but based on in-game results. 

Ready for testing: approved

PR #265, opened 2026-08-16
Integrate Faster Universalis

MichaHelder:
**Faster Universalis integrated** 
- To be specific, the 24h tick versions, aka, 1 tick = 1 day now 
- Ported over performance Defines 
- Ported over map market optimizations (actually cache map markers) 
- Remove all Societies of Pops at game initialization (as they don't do anything atm) 
- NOT (yet) ported: character culling, region removal 
- All Credits for these changes should go the authors of Faster Universalis, with a specific mention to Conner, its lead developer

Repository of Faster Universalis from which I ported the changes: https://github.com/conner-olsen/eu5-faster-universalis


cheetawilliam (approved):
tested it for a few hours, seems stable.

PR #267, opened 2026-08-18
Feature/inquisitor

Sun-Wu:
Added Invite Inquisitor Religious Action, which invites a cleric with low ADM, mid DIP, high MIL, and theologian trait, plus disallowed from leading armies or navies

### DIFF
--- a/in_game/common/generic_actions/MnT_catholic.txt
+++ b/in_game/common/generic_actions/MnT_catholic.txt
@@ -27,6 +27,7 @@
 				value = estate_satisfaction_severe_penalty
 				multiply = -1
 			}
+			NOT = { has_policy = council_of_three_lands_policy }
 		}
 		custom_tooltip = {
 			text = invite_inquisitor_no_countries_tt
@@ -40,6 +41,8 @@
 			}
 		}
 	}
+	
+	price = price:invite_foreign_cleric
 
 #	select_trigger = {
 #		looking_for_a = religion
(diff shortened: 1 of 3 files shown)


in_game/common/generic_actions/MnT_catholic.txt
```diff
@@ -0,0 +1,117 @@
...
+	automation_tick_frequency = 6
+
+	automation_tick = monthly
+	automation_tick_frequency = 12
+	player_automated_category = religiousdoctrines
+
+	sound = UI_action_religion_generic
+
+	potential = {
+		scope:actor.religion = religion:catholic
```
MichaHelder: Could you make this ?= instead of =

The reason is that sometimes tags have no religion, causing such an interaction to spam errors

in_game/common/generic_actions/MnT_catholic.txt
```diff
@@ -0,0 +1,117 @@
...
+	allow = {
+		papacy_active = yes
+		country_exists = c:PAP
+		scope:actor = {
+			has_law = inquisition_law
+			estate_type_allowed_in_cabinet = { estate_type = estate_type:clergy_estate }
+			estate_satisfaction:clergy_estate >= {
+				value = estate_satisfaction_severe_penalty
+				multiply = -1
+			}
```
MichaHelder, line 29: Isn't it much easier to just check for some defined value rather than the reversed severe penalty

in_game/common/generic_actions/MnT_catholic.txt
```diff
@@ -0,0 +1,117 @@
...
+			NOT = { has_policy = council_of_three_lands_policy }
+		}
+		custom_tooltip = {
```
MichaHelder: Might I ask why we check for another Catholic country to exist, whilst at the same time we are creating a new character in the scope:actor? If we are inviting an Inquisitor from scope:recipient, shouldn't the creation then also technically happen in the scope:recipient?

in_game/common/generic_actions/MnT_catholic.txt
Sun-Wu: I copied it from the Invite Foreign Cleric religious action, that's why it is like that.

in_game/common/generic_actions/MnT_catholic.txt
Sun-Wu: I had copied it from the Invite Foreign Cleric interaction and deleted it, still works fine.

MichaHelder:
Your branch currently produces the following three errors on startup: 

```
[09:35:19][interaction_target.cpp:3875]: Targetted action invite_inquisitor doesn't specify a recipient for the action
[09:35:19][message_handler.cpp:421]: Failed to find message type: PERFORM_invite_inquisitor_ACTION.
[09:35:19][generic_action_ai_list.cpp:82]: Action invite_inquisitor is not explicitly listed in an ai list! This has performance considerations!
```

Let me know if you want me to help fixing any of these

PR #268, opened 2026-08-20
Fix climate values

cheetawilliam:
fixed typo


Sun-Wu (approved):
Everything seems in order

PR #269, opened 2026-08-23
Fix small visibility bug with Great Pestilence and disable selling works of art (too exploitable)

MichaHelder:


cheetawilliam (approved):
fix and good tweak, lets hope tinto does their own rebalance in time.

PR #271, opened 2026-08-27
Feature/population

Malorn-Deslor:
The Population rework is ready for dev.

It may need some tweaking around certain areas, but the soft cap makes it more forgiving than before. Also population data seems to indicate that areas that are currently over pop capacity were historically areas that saw almost no population growth for several hundred years. Ergo...they SHOULD be at population capacity.


Documentation/Change log.md
```diff
@@ -68,6 +68,7 @@ Date: 21/05/2026
 - AI now has better non-royal marriage logic so that dynasties don't die out nearly as often due to refusal to wed
 - Development now directly reduced by looting and sieges.
 - Development will now decay at low enough prosperity levels.
+- Promotion rules changed, peasants unable to promote to nobles, many other classes able to promote 'upward'.
```
MichaHelder, line 71: Could you please add the changelog at the bottom of the file, patch notes for mod updates are created from the last entry (always one number ahead of current patch), so it will be a hassle to properly inform the players if you put it at 0.1.0 entry :P

Documentation/Change log.md
```diff
@@ -193,6 +194,8 @@ Date: 21/05/2026
 - Tribesmen pops now have same share of the tax base as peasants
 - Stop randomly spawning Eunuch children
 - Added tribes to the Steppe
+- Entire max population system reworked to be more climate related, and lower in general.
```
MichaHelder, line 197: Same here as above

MichaHelder (approved):
Discussed, playtested to satisfaction and the error log is good. 

Approved

PR #279, opened 2026-09-06
Fix building caps and rank gates left inconsistent after the RGO replacement

CBDecypher:
Five load-time error classes, all downstream of replacing RGOs with buildings, plus cleanup of the abandoned `land` good. Every number here was verified against a headless run. The over-cap errors this PR targets went from **34 to 2**, and the whole error log from **564 to 278-284** - the range because two vanilla error classes (`royal_marriage`, `ruler_term`) are written at runtime and their counts vary between identical sessions.

## Building caps

**Two caps were lost.** `japanese_clan_building_cap` and the stackable estate cap read RGO size, which is now zero everywhere, so they collapsed to their base. Restored as `REPLACE:` reading development and population instead.

**Five conversions ignored their own cap.** `destroy_invalid_buildings` converts the old buildings into `RGO_building_*`, carrying the level across, but the caps are fractional script values while building levels are integers — so a location at the cap converted to a level above it. Each conversion is now clamped:

```
change_building_level_in_location = {
    building = building_type:RGO_building_fruit
    value = {
        add = "location_building_level(building_type:fruit_orchard)"
        max = rgo_fruit_max_level
        floor = yes	#the caps are fractional; round down rather than over
    }
}
```

**The rural cap base was too low.** Base 1 could not hold the levels the base game itself places at setup, so freshly-loaded locations were immediately over cap. Raised to 3.

## Rank upgrade gates

`can_upgrade_country_rank_{empire,kingdom,duchy}` each required `prestige > 0`. Every country starts at 20 prestige, so all three gates were satisfied on day one by everyone — **251 county-rank countries qualified for a duchy upgrade at game start**. Raised to `> 25`, which is vanilla's own duchy bar. Vanilla asks 70 and 50 for empire and kingdom; those stay well below vanilla but now actually gate.

## Charcoal

`char_coal_maintenance` produced coal from clay alone, with no wood — a charcoal mound is packed with clay, but the coal comes from the wood. With no feedstock it returned 1900% profit against a declared target of 60%.

`debug_max_profit` is a **target with a 1% tolerance**, not a ceiling, so the lumber is solved for the declared target rather than copied from vanilla's ratio:

```
char_coal_maintenance:      lumber 0.3833, clay 0.10  -> coal 0.5   (target 60%)
improved_char_coal_wooded:  lumber 0.5056, clay 0.15  -> coal 0.75  (target 80%)
```

Both now sit inside tolerance.

## The `land` good

`land` was added as a base requirement for land-use RGOs and then abandoned — the good definition itself is commented out, so it never loaded. Left behind were 76 commented inputs, trade bans, modifier registrations, icon mappings, a colour file and unused localization, which together read as if restoring it was still planned.

Removed, plus the 39 orphaned `*_land` production-method identifiers whose suffix named an input that no longer exists. **Method identifiers only** — no recipe, quantity or balance value changes with them. No player-visible effect.

## Verification

Each change was checked by a headless run rather than by reading. A vanilla baseline run with the mod disabled logs 50 error lines against this tree's original 564, which is what established these were the mod's own errors rather than the base game's.


Malorn-Deslor:
Charcoal and rank are intentional choices, please explain in more detail what you mean in Building caps.

CBDecypher:
You zero RGO capacity globally in MnT_location.txt  REPLACE:location_base_values sets local_max_rgo_size = 0 and the development, population and rural-development contributions are commented out, which is all four of vanilla's additive sources.
Vanilla's building_caps.txt reads max_rgo_workers in four places - plantation_cap, rural_building_cap, estate_building_stackable_level, japanese_clan_building_cap.
You reauthored the first two. The other two weren't, so they multiply by zero.
I applied the same multiplier substitution the mod already uses, nothing new.

destroy_invalid_buildings.1 swaps each old building for its new RGO equivalent and gives the new one whatever level the old one had, so a lvl4 fruit_orchard becomes a lvl4 RGO_building_fruit.
Nothing checks whether the new building is allowed to be that big, and the two don't share a cap, so a level that was fine for the orchard can be over the limit for the building. So that should be clamped to the new building cap.
It needs floor = yes because these caps are fractional.
rgo_fruit_max_level starts at 0.5 and climbs with development so clamping to a cap of 2.6 without rounding down leaves it over again.

For charcoal, you removed wood for some reason and left 0.1 clay as requirement.
But you also did debug_max_profit=0.6 as a literal, where vanilla has it as a variable. Clay is 0.25 and coal is 1, so 0.1 clay costs 0.025 and makes 0.5 worth of coal. The engine says "too high profit at 1900% should be 60.00%". To hit 0.6 you'd need to spend 0.3125, which is 1.25 clay, 12.5x what's written.

Rank is correct in implementation and that's wrong on my end, sorry.

PR #281, opened 2026-09-09
Fix colonial integration

Mazutaki:
Colonial locations have not been integrating. Reason beeing: Only the integration status "conquered" made a location be regarded as "to be integrated".

Now, a scripted_trigger has been added that checks against the status "conquered" and "colonized" and is reused where needed, so should these requirements change ever again, there is _one_ code point to edit.


MichaHelder (approved):
Good bug fix, approved

PR #282, opened 2026-09-11
feat: show centers of importance on the location picture

mak8427:
Centers of Importance now show as tiles on the location picture (top left, one per category, tinted), greyed out when the location holds no center of that category.

Hovering a tile gives the center's effects, its score against the threshold for the next tier and a progress bar; hovering the score line opens the breakdown of that score, and every tooltip title opens the category's full tier progression.

Note: the 16 tile textures (`gfx/interface/icons/centers/<category>_<tier>.dds`) are still being drawn, so they are not part of this PR — the tiles render without icons until those land.

Also fixes the production score erroring out when a location has no urban goods value yet; the same bug blanked the centers mapmode tooltip.


MichaHelder (approved):
Looks good, I only moved the changelog down

PR #284, opened 2026-09-13
Fix/epbm ai countries get proper discounts

mikejaklitsch:
AI countries were not getting building maintenance discounts from estates: fixed
Foreign buildings were being calculated as having 0 maintenance, so building based countries got no reductions: fixed

Triggers to enter epbm effects were checking the wrong thing and should now be more accurate. any_active_estate was checking if any nobles estate existed in the entire world for instance and is_real_country could be false on rebel countries with land.

AI countries compare economic base to expenses to determine if they should go bankrupt. Day 1 this ratio was absurdly bad because RGOs are not staffed, and building maintenance numbers were not adjusted by the estate auto modifiers yet. I added a new setup file that gives every country a 2 month duration -100% building maintenance multiplier. on_game_start was able to apply the modifier before the bankruptcy, but the effect didn't take until the first month tick so a change to setup was the only way to resolve this.

Location wealth and wages branch has more complete changes to the epbm system that should resolve some more inefficiencies but this felt important enough to get out separately.


Malorn-Deslor (approved):
Glanced through the code and discussed reasoning for change. This is a critical fix. Approved.

PR #286, opened 2026-09-15
Diversify prices to build/upgrade Buildings

MichaHelder:
All RGO building types plus Amenities were using expand farming cost. 

Diversify RGO buildings to all available expand RGO price types, and create a new price type for Urban Amenities


RelMayers:
Hello, I have done not extensive check on upgrade types, those are differences between vanilla (source wiki sadly as I'm not at my PC right now):
Amber - gathering 
Beeswax - farming 
Clay - gathering 
Salt - gathering? Probably need split for coastal/lake based and mines
Sand - gathering

MichaHelder:
Its a bit dubious what all of them should be, there are only 5 categories whilst methods to get the materials vary wildly. 

E.g. Amber is usually mined, but sometimes kinda picked up? https://en.wikipedia.org/wiki/Amber#Extraction_and_processing

Beekeeping isn't really farming, but its not exactly gathering in the wild either. Its not entirely clear what the 'right' type of resource extraction should be in many cases. 

I'll take a look at salt, thats a good point. Evaporative extraction from the sea is more like gathering, but the RGO type is specific to places that really mined them. 

Edit, already set salt to mining, that is correct for the RGO building here. Evaporative salt collection is the salt_collector building. 

Clay and sand are really digged, neither mined nor collected. Unsure what would be best there, doubt it matters much tbh.

RelMayers:
open-pit mining for Amber it seems.
https://en.wikipedia.org/wiki/Amber_Coast
That does fit then nicely together with clay and sand as it's same type of open pit extraction.
Beekeeping is more similar to livestock(farming) than to fishing.

MichaHelder:
Updated beekeeping to be farming, and merged latest develop into this branch. Should be good for a merge

mak8427:
Looks good to me

Malorn-Deslor (approved):
Seems solid.

PR #288, opened 2026-09-21
Fix Levy Scales not working

MichaHelder:
- I have no idea why this fixed it
- But I confirmed it didn't work before
- And that it works now

Before: 
Schermafbeelding 2026-09-21 112122

After: 
Schermafbeelding 2026-09-21 112940


mikejaklitsch (approved):
I’m a sucker for before and after pics

Mazutaki:
> I have no idea why this fixed it 

https://github.com/Shemkent/EU5-Annotation-Modding/blob/main/annotations/script_values.md#modding-notes

Seems those values do not support REPLACE/INJECT, and are more comparable to defines.

